PR #65, opened 2020-07-07
Add free calls to match allocates and TEXT calls

domesticmouse:
Hey Tim,

Looking through the  `win32` package, and I'm guessing this sample would actually leak. Can you confirm if I am reading the library correctly, or if I'm missing something obvious?

brett

### DIFF
--- a/example/hello.dart
+++ b/example/hello.dart
@@ -83,8 +83,4 @@ void main() {
     TranslateMessage(msg.addressOf);
     DispatchMessage(msg.addressOf);
   }
-
-  free(CLASS_NAME);
-  free(wc.addressOf);
-  free(msg.addressOf);
 }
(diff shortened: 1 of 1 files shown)


timsneath:
Thank you!

And you didn't miss anything, I did :) This was one of the very first samples I wrote and I was far too lazy about deallocation at the time: I was just delighted to see pixels on the screen! I need to go through all the samples and make sure I'm not leaking in other places.

It's arguable, but I decided to skip the final `free()` calls at the end of `main()`, since Windows will automatically take care of freeing any remaining heap allocations when the process exits. The alternative is to add a bunch of extra boilerplate to do it properly (e.g. deallocating if an exception is thrown), which I don't think adds to the sample. Let me know if you disagree though, and we can hash it out.

domesticmouse:
Glad to know I wasn't misunderstanding what the library was doing. About the only thing I'd think about doing is adding a comment to the effect that we aren't freeing things because Windows will dealloc everything at program exit. Hopefully we can do something better here once the object finalization hook in https://github.com/dart-lang/sdk/issues/35770 is exposed to dart land.

PR #69, opened 2020-07-10
Investigating memory management for win32

domesticmouse:
Hey Tim,

I've started investigating how to wrap the win32 native API in a way that handles memory management issues, and improves ergonomics of using the win32 API. 

Would this make more sense as something bundled into the win32 package, or as a separate package?

brett

### DIFF
--- a/lib/with_win32.dart
+++ b/lib/with_win32.dart
@@ -1,0 +1,155 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:ffi';
+import 'package:ffi/ffi.dart';
+import 'win32.dart' as win32;
+
+export 'win32.dart'
+    hide
+        CreateWindowEx,
+        DispatchMessage,
+        DrawText,
+        GetClientRect,
+        GetMessage,
+        RegisterClass,
+        TranslateMessage;
+
+T withMSG<T>(T Function(win32.MSG) func) {
+  final msg = win32.MSG.allocate();
+  T t;
+  try {
+    t = func(msg);
+  } finally {
+    free(msg.addressOf);
+  }
+  return t;
+}
+
+T withPAINTSTRUCT<T>(T Function(win32.PAINTSTRUCT) func) {
+  final ps = win32.PAINTSTRUCT.allocate();
+  T t;
+  try {
+    t = func(ps);
+  } finally {
+    free(ps.addressOf);
+  }
+  return t;
+}
+
+T withRECT<T>(T Function(win32.RECT) func) {
+  final rect = win32.RECT.allocate();
+  T t;
+  try {
+    t = func(rect);
+  } finally {
+    free(rect.addressOf);
+  }
+  return t;
+}
+
+T withTEXT<T>(String text, T Function(Pointer<Utf16>) func) {
+  final utf16Text = win32.TEXT(text);
+  T t;
+  try {
+    t = func(utf16Text);
+  } finally {
+    free(utf16Text);
+  }
+  return t;
+}
+
+T withWNDCLASS<T>(T Function(win32.WNDCLASS) func) {
+  final wc = win32.WNDCLASS.allocate();
+  T t;
+  try {
+    t = func(wc);
+  } finally {
+    free(wc.addressOf);
+  }
+  return t;
+}
+
+T BeginPaint<T>(int hWnd, T Function(int hdc) func) => withPAINTSTRUCT((ps) {
+      final hdc = win32.BeginPaint(hWnd, ps.addressOf);
+      T t;
+      try {
+        t = func(hdc);
+      } finally {
+        win32.EndPaint(hWnd, ps.addressOf);
+      }
+      return t;
+    });
+
+int CreateWindowEx(
+        int dwExStyle,
+        String className,
+        String windowName,
+        int dwStyle,
+        int X,
+        int Y,
+        int nWidth,
+        int nHeight,
+        int hWndParent,
+        int hMenu,
+        int hInstance,
+        Pointer<Void> lpParam) =>
+    withTEXT(
+      className,
+      (className) => withTEXT(
+        windowName,
+        (windowName) {
+          var hWnd = win32.CreateWindowEx(
+            dwExStyle,
+            className,
+            windowName,
+            dwStyle,
+            X,
+            Y,
+            nWidth,
+            nHeight,
+            hWndParent,
+            hMenu,
+            hInstance,
+            lpParam,
+          );
+
+          if (hWnd == 0) {
+            final error = win32.GetLastError();
+            throw win32.WindowsException(win32.HRESULT_FROM_WIN32(error));
+          }
+
+          return hWnd;
+        },
+      ),
+    );
+
+int DispatchMessage(win32.MSG msg) => win32.DispatchMessage(msg.addressOf);
+
+int DrawText(int hdc, String text, int cchText, win32.RECT rect, int format) =>
+    withTEXT(
+        text,
+        (utf16text) =>
+            win32.DrawText(hdc, utf16text, cchText, rect.addressOf, format));
+
+T GetClientRect<T>(int hWnd, T Function(win32.RECT rect) func) =>
+    withRECT((rect) {
+      var result = win32.GetClientRect(hWnd, rect.addressOf);
+      if (result != 0) {
+        return func(rect);
+      } else {
+        final error = win32.GetLastError();
+        throw win32.WindowsException(win32.HRESULT_FROM_WIN32(error));
+      }
+    });
+
+int GetMessage(win32.MSG msg, int hWnd, int wMsgFilterMin, int wMsgFilterMax) =>
+    win32.GetMessage(msg.addressOf, hWnd, wMsgFilterMax, wMsgFilterMax);
+
+int RegisterClass(void Function(win32.WNDCLASS wc) func) => withWNDCLASS((wc) {
+      func(wc);
+      return win32.RegisterClass(wc.addressOf);
+    });
+
+int TranslateMessage(win32.MSG msg) => win32.TranslateMessage(msg.addressOf);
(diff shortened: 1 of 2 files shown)


timsneath:
This is really interesting -- thanks for sharing! 

I saw your earlier prototype and was starting to reply, but then I saw you'd pulled it back to work on this version.

A couple of thoughts:

1. Have you connected with Daco Harkes on his FFI finalizer work? Won't this provide a mechanism to allow these variables to be automatically garbage collected when they go out of scope? 
2. I've been thinking more generally about high-level APIs for Win32 -- I'm not sure yet what the right level of abstraction is.
  - Build high-level APIs that abstract Win32 away? Or make working with Win32 easier? I'm not sure we _really_ are aiming at people building traditional Win32 message-based apps here -- it's just a great proof-point. 
  - Create a separate plugin for each use case (paths, calendars, bluetooth, registry etc.) or keep them together? The former fits the Flutter federated model, the latter is better for a Windows-centric app.

Curious on your opinions :)

domesticmouse:
On Daco's Dart FFI Finalizer work, I am tracking https://github.com/dart-lang/sdk/issues/35770 and once a finalizer is exposed to Dart, all the `withXXX()` struct lifetime wrappers in this library will no longer be required. However, there are still cases in Win32 where we have things like BeginPaint/EndPaint where we can improve the Win32 API with the power of GC and higher order functions.

On building high level APIs vs making Win32 easier to use, I'd go the latter initially, as it is an easier goal to hit, and it's likely most Flutter developers targeting Windows are going to already know the specific API they want to use.

On the plugins front, I'd go down the single universal win32 plugin route, because Dart does treeshaking for the production build. There isn't a run time cost for unused Dart code in production, IIUC. As a bonus, the overall cost of maintaining one plugin is less than the cost of maintaining dozens.

=)

lib/with_win32.dart
```diff
@@ -0,0 +1,155 @@
...
+    withTEXT(
+        text,
+        (utf16text) =>
+            win32.DrawText(hdc, utf16text, cchText, rect.addressOf, format));
+
+T GetClientRect<T>(int hWnd, T Function(win32.RECT rect) func) =>
+    withRECT((rect) {
+      var result = win32.GetClientRect(hWnd, rect.addressOf);
+      if (result != 0) {
+        return func(rect);
```
timsneath, line 140: Does the `finally` block of a `try/finally` execute even when there's a `return`? I could only find reference to it firing regardless of whether an exception is thrown or not.

lib/with_win32.dart
```diff
@@ -0,0 +1,155 @@
...
+        t = func(hdc);
+      } finally {
+        win32.EndPaint(hWnd, ps.addressOf);
+      }
+      return t;
+    });
+
+int CreateWindowEx(
+        int dwExStyle,
+        String className,
```
timsneath, line 87: My understanding is that this `LPCTSTR` must be the same one as created through `RegisterClass`.  I haven't tried allocating the same string with a different pointer address, but normally Win32 apps hold a reference to this string and pass it through.

lib/with_win32.dart
```diff
@@ -0,0 +1,155 @@
...
+  T t;
+  try {
+    t = func(ps);
+  } finally {
+    free(ps.addressOf);
+  }
+  return t;
+}
+
+T withRECT<T>(T Function(win32.RECT) func) {
```
timsneath, line 41: This seems like it might get expensive, since `RECT`s are frequently passed from function to function.

lib/with_win32.dart
```diff
@@ -0,0 +1,155 @@
...
+T GetClientRect<T>(int hWnd, T Function(win32.RECT rect) func) =>
+    withRECT((rect) {
+      var result = win32.GetClientRect(hWnd, rect.addressOf);
+      if (result != 0) {
+        return func(rect);
+      } else {
+        final error = win32.GetLastError();
+        throw win32.WindowsException(win32.HRESULT_FROM_WIN32(error));
+      }
+    });
```
timsneath, line 145: In `GetClientRect()`, as with many Win32 functions, `rect` is a `Pointer` so that it can be used as an `out` value. Here I think what you return will be the same as what you input, no?

timsneath:
> On Daco's Dart FFI Finalizer work, I am tracking dart-lang/sdk#35770 and once a finalizer is exposed to Dart, all the withXXX() struct lifetime wrappers in this library will no longer be required. However, there are still cases in Win32 where we have things like BeginPaint/EndPaint where we can improve the Win32 API with the power of GC and higher order functions.

I'm all for making the Win32 projection more idiomatic Dart, so long as it doesn't change the signature or calling code in unpredictable ways. I've made some comments inline for the code itself where I wasn't clear whether that is so.

> On building high level APIs vs making Win32 easier to use, I'd go the latter initially, as it is an easier goal to hit, and it's likely most Flutter developers targeting Windows are going to already know the specific API they want to use.

Totally agree with your latter comment about target audiences. From the experiences I and others have had to date, the value of this is being able to take C sample code and rapidly convert it to the Dart equivalent so that it can be used from a platform plugin or other similar code. But the span of the Win32 API is so vast that this package will never cover all the bases, nor do we want to have to re-document the WIn32 API :)

> On the plugins front, I'd go down the single universal win32 plugin route, because Dart does treeshaking for the production build. There isn't a run time cost for unused Dart code in production, IIUC. As a bonus, the overall cost of maintaining one plugin is less than the cost of maintaining dozens.

Here I'm less clear. Because if the likes of `path_provider` are going to use this code, they will need a federated plugin for Windows (i.e. `path_provider_windows`). Or are you suggesting that the federated plugin itself comes from win32? 

There are some higher-level constructs that do wind up being useful in client plugin packages though. For example, it doesn't make sense to write [this wallpaper code](https://github.com/timsneath/filepicker_windows/blob/master/example/wallpaper/lib/wallpaper.dart) in the file_picker package when it could be available in a common location, which would likely be win32. Perhaps there's a separate high-level library within the package to distinguish this kind of code from the low-level C wrappers though?

domesticmouse:
> I'm all for making the Win32 projection more idiomatic Dart, so long as it doesn't change the signature or calling code in unpredictable ways. I've made some comments inline for the code itself where I wasn't clear whether that is so.

Fair enough, I'll work through the individual review comments later in the day.

> Totally agree with your latter comment about target audiences. From the experiences I and others have had to date, the value of this is being able to take C sample code and rapidly convert it to the Dart equivalent so that it can be used from a platform plugin or other similar code. But the span of the Win32 API is so vast that this package will never cover all the bases, nor do we want to have to re-document the WIn32 API :)

True enough.

> > On the plugins front, I'd go down the single universal win32 plugin route, because Dart does treeshaking for the production build. There isn't a run time cost for unused Dart code in production, IIUC. As a bonus, the overall cost of maintaining one plugin is less than the cost of maintaining dozens.
> 
> Here I'm less clear. Because if the likes of `path_provider` are going to use this code, they will need a federated plugin for Windows (i.e. `path_provider_windows`). Or are you suggesting that the federated plugin itself comes from win32?

I'm very unclear on how federated plugins can utilise `win32`, because my, potentially incorrect, understanding of plugin federation is that it happens at the point of the cross language channel. We'd need to kick in the win32 federated `path_provider` well before things are being marshalled into the channel to get the full value of the `win32` package.

> There are some higher-level constructs that do wind up being useful in client plugin packages though. For example, it doesn't make sense to write [this wallpaper code](https://github.com/timsneath/filepicker_windows/blob/master/example/wallpaper/lib/wallpaper.dart) in the file_picker package when it could be available in a common location, which would likely be win32. Perhaps there's a separate high-level library within the package to distinguish this kind of code from the low-level C wrappers though?

I totally agree with the separation of low level C wrappers and helper code, be it in the `win32` package or a sibling package.

lib/with_win32.dart
domesticmouse: Feel free to correct me here, but my understanding of how the Win32 C API works is that API calls expected structs to be passed in so that they aren't responsible for `malloc` & `free` of said structures. My understanding of `out` values was that it was introduced as a way of enabling multiple return values, and it was implemented via a double pointer indirection. As these functions are just taking a pointer to a `rect`, then there is no `out` magic going on here.

Remember, this is C, and immutable structs aren't a thing. =)

lib/with_win32.dart
domesticmouse: These `withXXX` functions are an attempt at making the lifetimes explicit, ala Rust and Java's try-with-resource. One of the wonderful things about Dart is the strength of it's GC implementation, as shown by Flutter's near blatant disregard for creating objects on each frame to describe the configuration of content on screen. So, I'm more worried about people `malloc`ing without matching `free` and thus slowly leaking an app to death, instead of chewing CPU.

IMNSHO, desktop class machines and GC are basically a solved problem for most non real time use cases. Look at C#. =)

lib/with_win32.dart
domesticmouse: That would be weird. I understand why C/C++ developers create a string once and hold onto it for the lifetime of the code, their language basically teaches people to be afraid of `malloc` and `free`. Most of the time it'd be a static string anyway, and thus packed into the PE, and not generated at run time.

With that said, I'd find it odd for the `CreateWindowEx` to only do pointer comparison, as opposed to string comparison. Most libc string comparison implementations I've seen have a pointer comparison fast path anyway, so it's a minimal optimisation that would break the explicit API contract. Is it documented as required to be pointer identical somewhere? The [CreateWindowEx docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexa) make no mention of pointer equality as a test.

This code appears to work as is, and IIUC this is also how the Rust team are approaching Win32 interop.

lib/with_win32.dart
domesticmouse: I expect so, given that is how `try/finally` works in other languages. Time to ask a language lawyer?

domesticmouse:
Thinking things through, I think this PR is probably best closed out. It was fun to experiment with re-designing the Win32 api to take advantage of higher order functions, i feel my time is better spent adding to the coverage of the C Win32 API.

Can we set up a tracking bug with a list of APIs we need to implement, in priority order?

Issue #70, opened 2020-07-13
Add wincred.h


hpoul:
I am interested in using wincred.h: https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credwritea
to port [my plugin](https://github.com/authpass/biometric_storage) to windows.

Do you have any plans to add support for it - or how much work is it? I'm not completely sure how much magic is autogenerated and what has to be manually added :-)

domesticmouse:
Hey @hpoul,

I think you will find @timsneath's answer in https://github.com/timsneath/win32/issues/64#issuecomment-655266444 helpful on the autogeneration / hand coding front.

Do you have an idea of which specific API calls you want to make?

hpoul:
@domesticmouse thanks for your fast reply, i think right now it's [CredWrite](https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credwritea), [CredRead](https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credreada), [CredDelete](https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creddeletea). I don't have experience with win32 APIs (yet), but those seem to be the closest to keychain read/write/delete :) 

not sure if it's better to have a generic ffi bridge, or just use the specific C functions in my library (which is how i've implemented it for linux right now), but if it's easier to build on this package, it would be great ;-)

timsneath:
Yeah, easy enough to add those functions, and they seem useful to others also. I'll put a quick pull request together. 

As far as the latter question is concerned, you're obviously welcome to do what works best for you. But three reasons why you might consider a direct dependency on win32:

- With Dart tree shaking, you're not paying any additional 'tax' for the parts of win32 you're not using; only those typedefs that you actually use are compiled into your plugin;
- Obviously any improvements made to the Win32 package come for free (e.g. Dart finalizer support, or improvements to struct layout, when they arrive)
- If you ever choose to expose any of the credentials outside of your own package, it might be helpful if the Structs etc. are in a common package.

hpoul:
cool thanks, that would be great.

To be honest my question was actually if it would be easier for me to work out how to add those APIs to `win32` and make a PR vs. just doing it in my own plugin. If I just have to use the package, I'll certainly go for it 😅️

timsneath:
You are very welcome to do so, if you want to get familiar with these APIs. A couple of tips:

[...]

PR #72, opened 2020-07-16
Adding methods for accessing windows credential manager #70

hpoul:
Adds
* `CredWrite`
* `CredRead`
* `CredFree` (required after CredRead)
* `CredDelete`

One weird behavior with `GetLastError`: when I first check the result code and afterwards call `GetLastError` it is always zero:

```dart
  final result =
      CredRead(TEXT('nonExistingCredential'), CRED_TYPE_GENERIC, 0, credPointer);
  if (result != TRUE) {
    final error = GetLastError();
    print('Error: $error');
  }
```

This will always print `Error: 0`, but when i move the GetLastError before the `if`:

```dart
  final result =
      CredRead(TEXT('nonExistingCredential'), CRED_TYPE_GENERIC, 0, credPointer);
  final error = GetLastError();
  if (result != TRUE) {
    print('Error: $error');
  }
```

It will print: `Error: 1168` (which is `ERROR_NOT_FOUND`) - Maybe dart invokes some system calls in the `if`? Did you stumble upon this before? The current version in this PR uses the first variant, ie. is buggy.. but I think it is kind of strange to *always* call `GetLastError()` without checking whether it was an error before?

### DIFF
--- a/example/credentials.dart
+++ b/example/credentials.dart
@@ -1,0 +1,59 @@
+import 'dart:convert';
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:win32/win32.dart';
+
+void write({String credentialName, String userName, String password}) {
+  print('Writing $credentialName ...');
+  final examplePassword = utf8.encode(password);
+  final blob = allocate<Uint8>(count: examplePassword.length);
+  final blobBytes = blob.asTypedList(examplePassword.length);
+  blobBytes.setAll(0, examplePassword);
+
+  final credential = CREDENTIAL.allocate()
+    ..Type = CRED_TYPE_GENERIC
+    ..TargetName = TEXT(credentialName)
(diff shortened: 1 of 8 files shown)


timsneath:
Thank you so much for your contribution! I made some minor reordering changes just for consistency, but otherwise this was great. I've also seen your `GetLastError` issue -- I think you may be right as to the diagnosis. Need to figure out what's going on here -- I wonder if it's a debug-mode thing, perhaps? 

I've overhauled the way I handle `HRESULT` since you checked out the code, since it wasn't being treated consistently as a signed value. As a result, I made a minor change to the `const` values in your code to store them as signed values.

Lastly, I fixed a minor bug, which was that `CredFree` was being called on the `Pointer<Pointer<CREDENTIAL>>`, rather than the `Pointer<CREDENTIAL>`. That fixed the occasional crash that I was getting with the test and example.

Thanks again -- I really appreciate it!

PR #92, opened 2020-08-07
Implement EnumProcessModulesEx

Hexer10:
I needed `EnumProcessModulesEx` since I was using dart 64bit to list the process modules of a 32bit process, maybe someone else would use this


timsneath:
Perfect! Thanks so much for your contribution!

PR #95, opened 2020-08-10
Automated generation of Win32 APIs

### LINKED ISSUE #94: Autogenerate Win32 APIs
1. Create a function to extract the comments, methods, parameters from the existing Dart files and export into a CSV file
2. Create a function to generate the core files from the CSV file

Advantages:

- Resilience to changes in the FFI format (easy to update the latter function rather than editing code
- Some better error checking (can inspect the CSV file with standardized tools like Google Sheets for consistency, e.g. is a `DWORD` always converted into an `Uint32`?)
- May be able to autogenerate documentation if there are public sources for API descriptions (can we scrape something)? 
- Allows switch to a more documentable format (see `FillConsoleOutputCharacter` in the `integrated-ffi` branch) -- it doesn't matter so much if the code is less readable if it's autogenerated anyway.
- Opportunity to read metadata in the future if Win32 APIs are documented in this way

Disadvantages
- Initial cost

timsneath:
Fixes #94.


tool/win32/win32api.csv
```diff
@@ -0,0 +1,216 @@
+ApiName, NeutralApiName, DllLibrary, WindowsPrototype, ReturnNative, ReturnDart, ParamCount, Param1Native, Param1Dart
```
domesticmouse, line 1: How would you feel about conversion of this CSV into a JSON or YAML file? I suspect a JSON file would be more easily maintained over CSV...

timsneath: I'm not entirely averse to it -- but curious why you think JSON is easier to maintain? 

Right now, I'm seeing a ton of value in it being in CSV: in particular, the ability to load into Google Sheets and batch edit is extremely attractive.

domesticmouse: I was thinking about adding summary docstrings to the 200 odd lines and I was thinking that doing that in CSV would look a bit ugly. If you get value out it being in CSV, then I suspect I can get it to work as is.

timsneath: Cool! I see that the MSDN [win32 documentation](https://github.com/MicrosoftDocs/win32/blob/docs/LICENSE) is licensed under CC-BY-4.0, so you could theoretically scrape directly from there.

This is all an ephemeral format that is specified only by the code that loads it, so if you wanted to create a JSON file for just the documentation, it would be easy enough to load that during the generation phase and map across. It's relatively easy to transform between formats, obviously, so we don't have to make any hard and fast rules here. 

The value I'm getting out of CSV is the ability to see a columnar view of all the methods, do things like generate a query for all the types, etc. It's an order of magnitude faster than the old approach for adding new APIs. 

I could imagine automating the process further too -- in theory I'd like to have a simple Flutter app where you ⌘+V something like this:
```
HRESULT SHGetKnownFolderPath(
  REFKNOWNFOLDERID rfid,
  DWORD            dwFlags,
  HANDLE           hToken,
  PWSTR            *ppszPath
);
```
and it generates a new CSV row automatically.

PR #121, opened 2020-11-24
Add NtSuspendProcess & NtResumeProcess APIs

Merrit:
These APIs are undocumented, but apparently widely used by Microsoft & Windows products and have been around for quite some time.

With only the PID for a process one can quickly and efficiently suspend / resume that process.


timsneath:
I'm pondering a little whether it's a good plan to start adding undocumented APIs like this. It's a slightly dangerous road to go down, since it sets a precedent and there are other ways to do this (albeit a little unwieldy).

Is there a project where you're planning to use them?

Merrit:
Yes actually there is. If you'd rather not include undocumented APIs I totally get that.

If you'd like, once I have things set up I could let you know where my APIs are so you can use them / point others to it / whathaveyou.

timsneath:
These are useful, there's no question. But I'm also conscious that win32 is used even by core Flutter libraries like filepicker, so at minimum I want to special case this. Perhaps the best way to handle this might be to expose a second library that others can optionally import -- with something like:

```dart
import 'package:win32/undocumented.dart';
```

Sorry -- I'll keep musing on this and explore a little. I very much appreciate the pull request.

Merrit:
Yeah, that makes sense to me.

Take your time, if you decide on anything you'd like help with just let me know. If you decide against including these API calls that is cool too.

No worries :smiley:

timsneath:
I'm sorry -- I think I'm going to close this without merging it. We're working on a way to automatically generate APIs from [Microsoft's metadata project](https://github.com/microsoft/win32metadata), and so adding these two APIs now will be a problem later since they won't be found in the Microsoft metadata cache. 

Apologies -- I really do appreciate the contribution, and hope you understand the rationale.

Merrit:
No worries at all. Thanks for consider it. :)

PR #124, opened 2020-12-27
Added RegSetValueEx

bsutton:
as the title says.

### DIFF
--- a/lib/src/advapi32.dart
+++ b/lib/src/advapi32.dart
@@ -174,3 +174,44 @@ int RegQueryValueEx(
   return _RegQueryValueEx(
       hKey, lpValueName, lpReserved, lpType, lpData, lpcbData);
 }
+
+/// Sets the data and type of a specified value under a registry key.
+/// To ensure that any string values (REG_SZ,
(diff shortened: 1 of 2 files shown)


Hexer10:
Have a look at the CONTRIBUTE.md:
> To create a new API, don't edit the main library files themselves; these get
overwritten. Instead, edit tool\win32\win32api.csv and run
tool\generate.cmd to update the library files. This might seem a pain, but
it gives protection against any changes to dart:ffi, since the typedef syntax
can be easily recreated as necessary.


If you edit manually the library files when the code is generated again all the changes are lost

timsneath:
Thanks so much for your contribution.

As @Hexer10 notes, the way to submit a new API is to update `win32api.csv` and then run `generate.cmd`. 

Curious -- where are you getting your API method signature from? I see different parameters at https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regsetvalueexw (e.g. `Reserved`, not `lpReserved`, `dwType`, not `lpType`).

In any event, I've added this API to the latest build at master; when I publish this to pub.dev next (`-nullsafety.9`), you should be able to use this API in your apps. Thanks -- and feel free to submit new patches as you need them!

PR #133, opened 2021-01-08
Add support for nested structs and pass-by-val

timsneath:


lib/src/structs.dart
```diff
@@ -671,35 +671,17 @@ class ACCEL extends Struct {
...
     ..cbSize = sizeOf<MONITORINFO>()
-    ..rcMonitorLeft = 0
-    ..rcMonitorTop = 0
-    ..rcMonitorRight = 0
-    ..rcMonitorBottom = 0
-    ..rcWorkLeft = 0
-    ..rcWorkTop = 0
-    ..rcWorkRight = 0
-    ..rcWorkBottom = 0
+    ..rcMonitor = RECT.allocate()
```
timsneath: @dcharkes What's the right way to initialize these variables now that they're nested structs? Is this going to work? Or do I have to go through every field in the nested struct and initialize them manually?

dcharkes: `RECT.allocate()` would malloc it, copy by value, and discard the pointer leaking memory.

If you want to initialize everything to zero you might instead want to use `memset`, and wrap `allocate` so that it 0-initializes everything. `allocate<MONITORINFO>().ref` -> `allocate<MONITORINFO>()..memset(0, sizeOf<MONITORINFO>()).ref`

timsneath: `memset`! Why have I never seen this in our examples... that's really helpful. I'll update accordingly. Thank you!!!

timsneath: Wait -- there _is_ no `memset`! Are you thinking of the C `memset()` call, @dcharkes? Or is that in a different branch?

timsneath: Never mind. I realized that the `allocate()` doesn't do more than wrap the Win32 `HeapAlloc()` call, so I've replaced it with my own `zeroAllocate()` function that uses the `HEAP_ZERO_MEMORY` flag. I'm going to raise an issue suggesting we use this by default (along with `calloc` on the UNIX side).

PR #134, opened 2021-01-08
COM interface tests

### LINKED ISSUE #132: Add tests for COM interfaces
The `Pointer<void>` bug exposed that not all COM methods aren't proven to load. We should add some tests to exercise every one as part of the generation.

timsneath:
Add tests for COM interfaces


timsneath:
Fixes #132

PR #144, opened 2021-01-20
Add example from slightfoot of WaveOut*

timsneath:


timsneath:
Closing this now that we have a WASAPI example that is fully Dart (#422).

PR #177, opened 2021-03-13
Calculate COM vtable starts from metadata

### LINKED ISSUE #163: Generate COM vtable starts from metadata
Here: https://github.com/timsneath/win32/blob/3193732957f6ff602a6e1ec5d1293b7386ff0bc7/tool/metadata/com.dart#L16-L22

bonukai:
This calculates vtable starts from metadata (#163) by summing up number of methods of all parents.

I'm not sure if this is the right way to do this, but all results match with current values.


timsneath:
Awesome! Thank you!

PR #213, opened 2021-04-19
add the GlobalSize interface

arcticfox1919:
This function is required when manipulating the clipboard


timsneath:
Thanks so much for the pull request! I really appreciate it. 

Just in case you want to contribute more APIs in the future, Win32 APIs should be added to the `win32api.json` file, otherwise they get overwritten next time the generation script and a unit test won't be created. More information here: [`CONTRIBUTING.md`](https://github.com/timsneath/win32/blob/main/CONTRIBUTING.md). 

I've made the change for this one. Again, thank you!

PR #246, opened 2021-06-11
Support `Windows.Data.Xml.Dom`

Sunbreak:

### DIFF
--- a/lib/src/com/IXmlNodeList.dart
+++ b/lib/src/com/IXmlNodeList.dart
@@ -1,0 +1,66 @@
+// IXmlNodeList.dart
+
+// THIS FILE IS GENERATED AUTOMATICALLY AND SHOULD NOT BE EDITED DIRECTLY.
+
+// ignore_for_file: unused_import
+
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+
+import '../combase.dart';
+import '../constants.dart';
+import '../exceptions.dart';
+import '../macros.dart';
+import '../ole32.dart';
+import '../structs.dart';
+import '../structs.g.dart';
+import '../utils.dart';
+
+import '../winrt/winrt_constants.dart';
+
+/// @nodoc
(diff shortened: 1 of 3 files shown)


Sunbreak:
I'm porting https://github.com/mohabouje/WinToast to Dart/Flutter, which needs `Windows.Data.Xml.Dom` features

When add `Windows.Data.Xml.Dom` there're several problems, could you take a look? @timsneath

lib/src/com/IXmlNodeList.dart
```diff
@@ -0,0 +1,66 @@
...
+typedef _get_Length_Dart = int Function(Pointer obj, Pointer<Uint32> value);
+
+typedef _Item_Native = Int32 Function(
+    Pointer obj, Uint32 index, Pointer<Pointer> result);
+typedef _Item_Dart = int Function(
+    Pointer obj, int index, Pointer<Pointer> result);
+
+/// {@category Interface}
+/// {@category winrt}
+class IXmlNodeList {
```
Sunbreak, line 35: `IXmlNodeList` is a subtype of `IInspectable` but the generation omit the inheritance

@timsneath how could I dig inside? Or should I add the inheritance manually?

timsneath: It looks like `IXmlNodeList` implements `IReadOnlyList<IXmlNode>` and `IEnumerable<IXmlNode>`. I think my projection code is broken for WinRT interfaces with generics at present, so it doesn't know what to do. 

Curious though -- can't you use the Dart [xml](https://pub.dev/packages/xml) package to parse the XML? I think that's going to be easier...

timsneath: Opened #247.

Sunbreak: > Opened #247.

Thanks a lot

> 

> Curious though -- can't you use the Dart [xml](https://pub.dev/packages/xml) package to parse the XML? I think that's going to be easier...

`IToastNotificationManagerStatics`'s `GetTemplateContent` results in a `IXmlDocument`. So I guess I have to use `Windows.Data.Xml.Dom`

PR #250, opened 2021-06-11
Add helper class HString

### LINKED ISSUE #249: [Feature] Helper class/function for HSTRING
# Background

- UserDataPath

https://github.com/timsneath/win32/blob/a608e2e47c8c6f2cc94180b4de101592b308a54c/lib/src/com/UserDataPaths.dart#L415

https://github.com/timsneath/win32/blob/a608e2e47c8c6f2cc94180b4de101592b308a54c/lib/src/com/UserDataPaths.dart#L434-L435

```dart
  static UserDataPaths GetForUser(Pointer user) {
    final hClassName = convertToHString(_className);Tim Sneath, 2 months ago: • Move statics into UserDataPaths

      // ...
    } finally {
      WindowsDeleteString(hClassName.value);
      free(hClassName);
      // ...
    }
  }
```

- ToastNotification

https://github.com/Sunbreak/win32/blob/3c5085a97c6d9e20c098e442ba6b3ab32d5fe9d9/lib/src/com/ToastNotification.dart#L158

https://github.com/Sunbreak/win32/blob/3c5085a97c6d9e20c098e442ba6b3ab32d5fe9d9/lib/src/com/ToastNotification.dart#L179-L180

```dart
  static ToastNotification Create(Pointer content) {
    final hClassName = convertToHString(_className);

      // ...
    } finally {
      WindowsDeleteString(hClassName.value);
      free(hClassName);
      // ...
    }
  }
```

- IToastNotificationManagerStatics

`IToastNotificationManagerStatics` would need `HSTRING` for `RoGetActivationFactory` as well

# Request

Add helper class/function for `HSTRING` in case of omitting `WindowsDeleteString`

Sunbreak:
fix #249 


timsneath:
@Sunbreak Thanks again! 

I spent some time reading https://docs.microsoft.com/en-us/windows/win32/winrt/hstring, and I think I'd misunderstood the way that `HSTRING` works when I first implemented it, and that error has pervaded into the code. Per this page, "the HSTRING handle is a standard handle type" -- which means win32 should treat it as an `int` rather than a `Pointer<IntPtr>`. 

Furthermore, `WindowsDeleteString` handles the underlying memory allocation -- it manages a refcount and when the value hits 0 it deallocates the string. So, if I've read this correctly, the Dart code should _never_ free the underlying storage -- it should call `WindowsDeleteString` and then forget about it. 

The good news is that this simplifies things. I've adjusted your code accordingly to turn it back into an int, and have made a few other changes elsewhere to correspond.

Please take a look and let me know if you disagree with my (newly improved) understanding!

PR #256, opened 2021-06-17
Feature/storage structs

Sunbreak:
Add several `Windows.Win32.Storage.StructuredStorage` structs for further usage


timsneath:
Great addition -- thank you for the tidy pull request! I've fixed the namespace for `STATPROPSTG`, which was causing the wrong struct to be generated; otherwise this looks great. 

I'm _very_ appreciative of your contributions -- thank you for collaborating to make this better!

PR #337, opened 2021-12-22
Switch exclusion Lists for Sets for faster contains() checks

DanTup:
These are used quite a lot to check if an item should be excluded and switching to a `Set` (where `contains()` is `O(1)` instead of `O(n)` like in `List`) seems to make a noticable difference - at least on my machine (a Core i5-4430):

```
Before:
[0:09:59.595648] Projection generation completed.

After:
[0:08:58.361169] Projection generation completed.
```

### DIFF
--- a/tool/v3/exclusions.dart
+++ b/tool/v3/exclusions.dart
@@ -1,4 +1,4 @@
-const excludedNamespaces = <String>[
+const excludedNamespaces = <String>{
   // We're focusing on client APIs, rather than server-centric APIs like IIS.
   // We're also excluding APIs that are no longer well supported, like the old
   // Trident MSHTML engine.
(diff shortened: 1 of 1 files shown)


DanTup:
(@timsneath I don't know if this is the right branch to change this - I noticed the file was in a different place on the main branch and assumed you're working on improvements here so it made sense to add on here, but feel free to just make the change on main and close this if it's not the best place!)

timsneath:
Oh, that's a great idea. I totally agree, sets make complete sense here. 

Although it's an interesting potential compiler optimization for const Lists -- if we know that they don't contain duplicates, I wonder if we could convert them automatically to use a Set representation at compile time?

DanTup:
> I wonder if we could convert them automatically to use a Set representation at compile time?

I'm not too familiar with the possibilities here, though I suspect _changing_ the type would be difficult (it might be passed to functions expecting a `List` etc.), although I wonder if calls to `.contains()` could be redirected to a copy stored as a `Set`? Maybe it's a bit too "magic" though, I'm not sure.

Maybe there's value in a lint that reports calls to `contains()` on a `const List` that could easily be a `const Set`? I'm not sure if perf things are usually linted (although I do recall some on async io) or if there are times where this doesn't make sense to do (on small lists where it's called infrequently there's probably little to gain, although I'm not sure that would make it a _bad_ idea to use a `Set`)? 🤔

timsneath:
Yeah, I was thinking more of an internal representation change, rather than changing the type under the covers. 

I don't see any reason why we wouldn't lint though. [Here's a good example](https://dart.dev/tools/linter-rules#use_string_buffers) of a perf-only lint.

PR #342, opened 2022-01-12
Wslapi

ElMoribond:
Hello Tim

I would like to add the API calls to [wslapi.dll ](https://docs.microsoft.com/en-us/windows/win32/api/wslapi/)but I need help.

These functions have the expected behavior and therefore seem to be ok:
- WslIsDistributionRegistered
- WslRegisterDistribution
- WslUnregisterDistribution
- WslConfigureDistribution
- WslLaunchInteractive

Stay these 2 there:

- WslGetDistributionConfiguration
I'm sure I'm using the wrong type of pointer for defaultEnvironmentVariables.
I have tried with int derivated pointer but don't know how to iterate through the returned list.

- WslLaunch
When the API is called the stdout/stderr output is displayed in the console.
I don't think this is desirable and I don't know how to avoid it.
Actually I didn't understand how to use redirects.

Anyway, I need help.
I will correct the tests when I have answers to these problems.

Thanks, Mik


timsneath:
Mik, this is a great addition -- thank you! 

I've made some changes to this, as you can see, building on your work. I took your tests and turned them into an example, which includes retrieving configuration information and running a test command. I wasn't able to keep them as a test, because they make specific assumptions about the test environment that don't hold for CI/CD, unfortunately. 

The updated example is in `example\wsl.dart`. On my machine with a single Ubuntu WSL instance, it prints:

```
[C:\src\win32] dart example\wsl.dart
Distribution: Ubuntu
Version: 2
Windows drives automatically mounted: true
Environment variables:
HOSTTYPE=x86_64
LANG=en_US.UTF-8
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
TERM=xterm-256color
Test command (uname -a) reports:
Linux tim-surface 5.10.60.1-microsoft-standard-WSL2 #1 SMP Wed Aug 25 23:20:18 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Command returned exit code: 0
```

ElMoribond:
Just question, why do you change this [signature](https://github.com/timsneath/win32/blob/main/lib/src/api-ms-win-wsl-api-l1-1-0.dart#L98-L110)?

timsneath:
Hi, @ElMoribond -- thanks again for your contribution!

If you're talking about the change to the returned value, the Win32 signature is:

```c
BOOL WslIsDistributionRegistered(
  PCWSTR distributionName
);
```

In your original PR, it was:

```dart
bool WslIsDistributionRegistered(Pointer<Utf16> distributionName) =>
    _WslIsDistributionRegistered(distributionName);

late final _WslIsDistributionRegistered = _wslapi.lookupFunction<
    Bool Function(Pointer<Utf16> distributionName),
    bool Function(
        Pointer<Utf16> distributionName)>('WslIsDistributionRegistered');
```

I changed it (through the automatic generation tool) to:

```dart
int WslIsDistributionRegistered(Pointer<Utf16> distributionName) =>
    _WslIsDistributionRegistered(distributionName);

late final _WslIsDistributionRegistered =
    _api_ms_win_wsl_api_l1_1_0.lookupFunction<
        Int32 Function(Pointer<Utf16> distributionName),
        int Function(
            Pointer<Utf16> distributionName)>('WslIsDistributionRegistered');
```

The simple reason is that the Win32 `BOOL` is declared as an `int` in the Win32 headers (`minwindef.h`):

```c
typedef int                 BOOL;
```

which is an `Int32` in Dart. Win32 declares two macros `TRUE` and `FALSE` that represent the Boolean values.

I also changed this from a DLL reference ("`wslapi.dll`") to an API set reference, matching the Windows documentation [for this API](https://docs.microsoft.com/en-us/windows/win32/api/wslapi/nf-wslapi-wslisdistributionregistered). They might change the DLL name, but the API set is promised to remain stable.

Hope that's helpful! Please feel free to make future contributions... I greatly appreciate it!

ElMoribond:
It is I who thank you for having made this great package which makes life so much easier :+1:
And then my participation is modest, luckily you were there to correct.

Thank you also for your explanations (I should have thought about it myself:confused:).

PR #343, opened 2022-01-13
Add `SetEvent`

untp:
`CreateEvent`, `OpenEvent`, `ResetEvent` exists but `SetEvent` is missing.


timsneath:
That seems like an oversight -- thank you so much for the contribution!

Issue #349, opened 2022-01-21
Remove conflict with `Char` in `dart:ffi`


dcharkes:
https://dart-review.googlesource.com/c/sdk/+/228541 added a `Char` type to `dart:ffi`.

Unfortunately that breaks `package:win32` and in turn the Flutter build: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8824468064587445729/+/u/Android_Views_Integration_Tests/stdout

The easiest fix would be to just `import 'dart:ffi' hide Char;` in https://github.com/timsneath/win32/blob/main/lib/src/structs.g.dart.

(And then release a new version, and roll that into Flutter, and any packages that are used in Flutter and roll those into Flutter.)

timsneath:
@dcharkes This is fixed in win32 v2.3.8.

a-siva:
Can we close this issue ?

timsneath:
It's closed, no?

PR #362, opened 2022-02-09
Hide Char in preparation for new package:ffi

Sunbreak:

### DIFF
--- a/lib/src/structs.g.dart
+++ b/lib/src/structs.g.dart
@@ -47,7 +47,9 @@ import 'dart:ffi' hide Char;
 
 import 'dart:typed_data';
 
-import 'package:ffi/ffi.dart';
+// `package:ffi` 1.2.0-dev.0 will introduce a new `Char` FFI type.
+// ignore: undefined_hidden_name
+import 'package:ffi/ffi.dart' hide Char;
 
 import 'callbacks.dart';
 import 'com/IDispatch.dart';
--- a/tool/metadata/generate_win32_structs.dart
+++ b/tool/metadata/generate_win32_structs.dart
@@ -63,7 +63,9 @@ import 'dart:ffi' hide Char;
 
 import 'dart:typed_data';
 
-import 'package:ffi/ffi.dart';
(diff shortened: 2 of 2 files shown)


Sunbreak:
Since https://pub.dev/packages/ffi/versions/1.2.0-dev.0 `Char` is inside `package:ffi` according to https://github.com/dart-lang/ffi/pull/128

Related issue: https://github.com/timsneath/win32/issues/349

timsneath:
This would have been a great patch, but we actually decided to not ship `Char` in package:ffi and then move it to dart:ffi. We discovered we were causing a lot of name clashes, so we [retracted 1.2.0](https://github.com/dart-lang/ffi/issues/133) and are going to just ship the new `Char` data type in dart:ffi itself in Dart 2.17.0. 

As you note, this continues to cause a name clash for the prerelease version, but there are obviously no guarantees that prerelease code is bug-free. I may change the upper bound of win32's dependency on package:ffi to avoid version 1.2.0-dev.0, but even that doesn't solve for people who already have existing versions. 

@dcharkes should we retract 1.2.0-dev.0 as well as 1.2.0, just for completeness? I also saw [this isolated incident](https://github.com/flutter/flutter/issues/98088) in the Flutter issue database.

dcharkes:
> @dcharkes should we retract 1.2.0-dev.0 as well as 1.2.0, just for completeness?

ffigen's dev release depends on it, and people are experimenting with it and finding bugs. So, I'd say no.

> I also saw [this isolated incident](https://github.com/flutter/flutter/issues/98088) in the Flutter issue database.

Pub should never select pre-releases by itself. So maybe the bug reported upgraded his project right in the time 1.2.0 stable was live?

timsneath:
These types have landed in 2.17 (not yet released to stable), and they have been renamed to `UnsignedChar` and `SignedChar`, which solves the problem of name clashes altogether. See https://api.dart.dev/be/184424/dart-ffi/dart-ffi-library.html. 

As a result, I'm closing this PR -- hope that's OK?

PR #382, opened 2022-03-25
Hide GetLastError

Sunbreak:


timsneath:
I've sent a mail to the FFI engineer to see if it's possible to get this fixed in Dart. 

I'm less keen to hide it here, because if it gets fixed in Dart, users still won't be able to use it. But I'll flag it as an open bug at least.

timsneath:
Filed #384 as a public note.

timsneath:
Closing for now in favor of #384, which is on the roadmap for FFI. Better to have the code available, IMO, so that when it's implemented this starts working.

PR #390, opened 2022-03-27
Add setupapi example

Sunbreak:


timsneath:
@Sunbreak, as ever, thank you for your contribution!

Would it be possible to add some comments to this, either to the header or to the console output? I'm a bit confused what this does or how to know when it's successful. On my machine it shows:

```
net device 0x00000001
net device 0x00000002
...
hid interface 176
hid interface 226
hid interface 194
hid interface 226
...
```

I assume this is intended behavior, but I'm not familiar with the API that you're using, I'm afraid.

Sunbreak:
Added simple comments

Sunbreak:
A more specific sample: https://github.com/Sunbreak/logic_conf.dart/commit/446dc328d994d15989ebb55b64c62a5a9e9af233

PR #392, opened 2022-03-27
Add bluetoothle example

Sunbreak:


timsneath:
Filed https://github.com/microsoft/win32metadata/issues/856 to reflect that these methods are not correctly treating `HDEVINFO` as a `HANDLE`.

timsneath:
Thanks so much for the sample, @Sunbreak! I made a few tweaks for consistency with other examples and moved the constants into the main win32 package. I also left you two `TODO` comments to take a look at.

Sunbreak:
> Filed [microsoft/win32metadata#856](https://github.com/microsoft/win32metadata/issues/856) to reflect that these methods are not correctly treating `HDEVINFO` as a `HANDLE`.

Is it blocked?

Sunbreak:
>       // TODO(sunbreak): requiredSizePtr is in bytes, no? I think you're
>       // allocating twice the size you need.`

I'd like to re-check it

>  // TODO(sunbreak): if path was passed as a String, you could just call
>  // path.toNativeUtf16() to get yourself a Pointer<Utf16>.

I'm not sure it would be feasible to use `utf8.decode(deviceInterfaceDetailDataPtr.getDevicePathData(requiredSizePtr.value))` and `path.toNativeUtf16` on no-English environment

`Uint16List` would be more compatible

timsneath:
@Sunbreak, with regard to https://github.com/timsneath/win32/pull/392#issuecomment-1080170927, it's not blocked, but the signature will change when they correct the metadata. Instead of returning a `void *` which my projection converts to a `Pointer`, it will return an `IntPtr` / `int`. 

We just have to be aware that any code that relies on the current signature will break when this is fixed, so any code that depends on the current implementation should take a tight version dependency for the time being.

timsneath:
I'm a bit confused by this: https://github.com/timsneath/win32/pull/392#issuecomment-1080194869. Doesn't it return a struct containing a `WCHAR` ([SP_DEVICE_INTERFACE_DETAIL_DATA_W](https://docs.microsoft.com/en-us/windows/win32/api/setupapi/ns-setupapi-sp_device_interface_detail_data_w))? That is the same as a Dart `Utf16`, isn't it? 

Maybe I can help craft a version that you can test on a non-English machine? 

I've just published [win32 2.5.0](https://pub.dev/packages/win32/changelog), by the way, if you want to test things out. I included the setupapi functions, and cherry picked the `constants.dart` changes from this pull request. Thanks for your great contributions so far!

Sunbreak:
I'll re-check `WCHAR` later off work. Thanks for you advise

Sunbreak:
Fixed two TODOs. Could you review again?

timsneath:
Thank you so much! LGTM!

PR #393, opened 2022-03-28
added example reading out a monitors physical size

krjw-eyev:
I have added a small example on how to read out the physical size of a monitor in millimeters from EDID data.


example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+class Size {
+  double width;
+  double height;
+  Size(this.width, this.height);
+}
+
+Size getMonitorSizeInMM() {
+  final guidptr = calloc<GUID>();
+
+  guidptr.ref.setGUID(GUID_CLASS_MONITOR);
```
timsneath: You can use the helper function here to set the GUID: 

```suggestion
  final guidptr = GUIDFromString(GUID_CLASS_MONITOR);
```

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+    final hDevRegKey = SetupDiOpenDevRegKey(
+        ptr, data, DICS_FLAG_GLOBAL, 0, DIREG_DEV, KEY_READ);
+
+    var retValue = ERROR_SUCCESS;
+    const nameSize = 128;
+    final lpValueName = calloc<Uint16>(nameSize).cast<Utf16>();
+    const edidDataSize = 1024;
+
+    final lpcchValueName = calloc<DWORD>()..value = nameSize;
+    final lpType = calloc<DWORD>()..value = nameSize;
```
timsneath: `lpType` is an out parameter, so there's no need to set its value here. And since it's optional, you can just pass `nullptr` instead of allocating a variable here. (see https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regenumvaluew)

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  guidptr.ref.setGUID(GUID_CLASS_MONITOR);
+  // Get the handle for the first monitor.
+  final ptr = SetupDiGetClassDevs(guidptr, nullptr, 0, DIGCF_PRESENT);
+  var width = 0;
+  var height = 0;
+
+  final data = calloc<SP_DEVINFO_DATA>();
+  data.ref.cbSize = sizeOf<SP_DEVINFO_DATA>();
+  // Get the device information for the first member of the first moitor
```
timsneath: `moitor` -> `monitor`.

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+  final ret = SetupDiEnumDeviceInfo(ptr, 0, data);
+  if (ret == TRUE) {
+    // Get the registry key for the first member of the first moitor
```
timsneath: `moitor` -> `monitor`.

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+    final lpType = calloc<DWORD>()..value = nameSize;
+    final lpData = calloc<BYTE>(edidDataSize);
+    final lpcbData = calloc<DWORD>()..value = edidDataSize;
+
+    // Get the first value of the registry key for the first member of the first moitor
+    retValue = RegEnumValue(hDevRegKey, 0, lpValueName, lpcchValueName, nullptr,
+        lpType, lpData, lpcbData);
+
+    if (retValue == ERROR_SUCCESS && lpValueName.toDartString() == 'EDID') {
+      width = ((lpData[68] & 0xF0) << 4) + lpData[66];
```
timsneath: Do these numbers have special meaning? If so, you might consider a constant and comment to explain to other readers, e.g.

```dart
// size is stored across three bytes, in the following format: ___
const sizeIdx = 66;
...
width = (lpData[sizeIdx+2] & 0xF0 << 4) + lpData[sizeIdx];
```

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+  free(data);
+
+  SetupDiDestroyDeviceInfoList(ptr);
+  free(guidptr);
+  return Size(width.toDouble(), height.toDouble());
+}
+
+Size getMonitorSizeInMMBackup() {
+  final hwnd = calloc<HWND>();
+  final hdc = GetDC(hwnd.value);
```
timsneath: `HWND` is an `int` (that happens to be the same size as a pointer). It doesn't need allocating. If you want the DC for the screen, simply use `final hdc = GetDC(NULL);`.

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+    const nameSize = 128;
+    final lpValueName = calloc<Uint16>(nameSize).cast<Utf16>();
+    const edidDataSize = 1024;
+
+    final lpcchValueName = calloc<DWORD>()..value = nameSize;
+    final lpType = calloc<DWORD>()..value = nameSize;
+    final lpData = calloc<BYTE>(edidDataSize);
+    final lpcbData = calloc<DWORD>()..value = edidDataSize;
+
+    // Get the first value of the registry key for the first member of the first moitor
```
timsneath: Sorry :) `moitor` -> `monitor` again!

example/monitor_edid.dart
```diff
@@ -0,0 +1,88 @@
...
+  final data = calloc<SP_DEVINFO_DATA>();
+  data.ref.cbSize = sizeOf<SP_DEVINFO_DATA>();
+  // Get the device information for the first member of the first moitor
+  final ret = SetupDiEnumDeviceInfo(ptr, 0, data);
+  if (ret == TRUE) {
+    // Get the registry key for the first member of the first moitor
+    final hDevRegKey = SetupDiOpenDevRegKey(
+        ptr, data, DICS_FLAG_GLOBAL, 0, DIREG_DEV, KEY_READ);
+
+    var retValue = ERROR_SUCCESS;
```
timsneath: Maybe just use this with `final retValue = RegEnumValue(...)` later?

timsneath (changes requested):
Thanks so much for this PR! I've moved the constants over to the library for others to use, and added some review comments here to address.

example/monitor_edid.dart
krjw-eyev: they do I will have a look later.

example/monitor_edid.dart
krjw-eyev: I didn't know thanks

example/monitor_edid.dart
krjw-eyev: copy pasta sorry :)

krjw-eyev:
@timsneath I think I have fixed everything. If there is anything else please let me know.

krjw-eyev (commented):
lgtm

timsneath (approved):

PR #398, opened 2022-03-30
Two small typos in comments fixed.

gaddlord:


lib/src/constants_winsock.dart
```diff
@@ -16,7 +16,7 @@ const SOCK_STREAM = 1;
 
 /// Datagram socket.
 ///
-/// A socket type that supports datagrams, which are connectionless, unreliable
+/// A socket type that supports datagrams, which are connection-less, unreliable
```
timsneath: "connectionless" seems OK, I think? (For example: https://en.wikipedia.org/wiki/Connectionless_communication)

This text comes from the [Microsoft documentation](https://docs.microsoft.com/en-us/windows/win32/api/ws2def/ns-ws2def-csaddr_info).

timsneath:
Thanks! I'm not sure that the hyphen is correct for the former. 

The latter is a good catch -- I'll merge that if you don't mind reverting the other.

PR #405, opened 2022-04-15
Project `Pointer<COMObject>` types correctly

halildurmus:
I've been working on this for a few days and I think I figured it out (even though I don't fully understand how it works 😄). 

Before this PR, `ICalendar`'s `Clone()` would be projected as:
 ```dart
Pointer<COMObject> Clone() {
    final retValuePtr = calloc<Pointer<COMObject>>();

    try {
      final hr = ptr.ref.lpVtbl.value
          .elementAt(6)
          .cast<
              Pointer<
                  NativeFunction<
                      HRESULT Function(
            Pointer,
            Pointer<Pointer<COMObject>>,
          )>>>()
          .value
          .asFunction<
              int Function(
            Pointer,
            Pointer<Pointer<COMObject>>,
          )>()(
        ptr.ref.lpVtbl,
        retValuePtr,
      );

      final retValue = retValuePtr.value;
      return retValue;
    } finally {
      free(retValuePtr);
    }
  }
```
With this projection, `Clone()` successfully clones the `Calendar` object and `ICalendar.Compare()` compares the calendar objects correctly, _but_ you couldn't use the `Pointer<COMObject>` returned from `Clone()` in the `ICalendar` class.
When you tried to use it, the process would crash every time. e.g.:
```dart
    ...
    final calendar = ICalendar(comObject);
    final clonedCalendar = ICalendar(calendar.Clone());
    print(clonedCalendar.GetDateTime()); // Crashes here
    ...
```

With this PR, `Clone()` will be generated as:
 ```dart
  Pointer<COMObject> Clone() {
    final retValuePtr = calloc<COMObject>();

    final hr = ptr.ref.vtable
        .elementAt(6)
        .cast<
            Pointer<
                NativeFunction<
                    HRESULT Function(
          Pointer,
          Pointer<COMObject>,
        )>>>()
        .value
        .asFunction<
            int Function(
          Pointer,
          Pointer<COMObject>,
        )>()(
      ptr.ref.lpVtbl,
      retValuePtr,
    );

    return retValuePtr;
  }
```
With this projection, you will be able to successfully use the `Pointer<COMObject>` returned from `Clone()` in the `ICalendar` class.

Also, I had to modify the projection of the methods that take `Pointer<COMObject>` as parameters because the process would crash every time when I try to use these.

For example, `ICalendar`'s `Compare()`:
```diff
int Compare(
    Pointer<COMObject> other,
  ) {
    final retValuePtr = calloc<Int32>();

    try {
      final hr = ptr.ref.lpVtbl.value
          .elementAt(93)
          .cast<
              Pointer<
                  NativeFunction<
                      HRESULT Function(
            Pointer,
            Pointer<COMObject> other,
            Pointer<Int32>,
          )>>>()
          .value
          .asFunction<
              int Function(
            Pointer,
            Pointer<COMObject> other,
            Pointer<Int32>,
          )>()(
        ptr.ref.lpVtbl,
-       other,
+       Pointer<Pointer<COMObject>>.fromAddress(other.address).value,
        retValuePtr,
      );

      if (FAILED(hr)) throw WindowsException(hr);

      final retValue = retValuePtr.value;
      return retValue;
    } finally {
      free(retValuePtr);
    }
  }
```

As I said, I don't fully understand how it works and I'm not sure this is the best way to project these but I hope this will give you an idea.

This PR will also allow us to project types like `IVector` and `IVectorView` and much more in the future. I actually manually projected these two types and will send a PR later today.

### DIFF
--- a/example/calendar.dart
+++ b/example/calendar.dart
@@ -23,8 +23,11 @@ void main() {
     final comObject =
         CreateObject('Windows.Globalization.Calendar', IID_ICalendar);
     final calendar = ICalendar(comObject);
-
     print(calendarData(calendar));
+
+    final clonedCalendar = ICalendar(calendar.Clone());
+    print(
+        'Comparison result of calendar and its clone: ${clonedCalendar.Compare(calendar.ptr)}');
 
     calendar.ChangeCalendarSystem(japaneseCalendar);
     print(calendarData(calendar));
@@ -36,6 +39,7 @@ void main() {
     print(dateTime);
 
     free(comObject);
+    free(clonedCalendar.ptr);
   } finally {
     WindowsDeleteString(japaneseCalendar);
     WindowsDeleteString(hebrewCalendar);
--- a/lib/src/com/ICalendar.dart
+++ b/lib/src/com/ICalendar.dart
@@ -37,35 +37,30 @@ class ICalendar extends IInspectable {
   ICalendar(Pointer<COMObject> ptr) : super(ptr);
 
   Pointer<COMObject> Clone() {
-    final retValuePtr = calloc<Pointer<COMObject>>();
+    final retValuePtr = calloc<COMObject>();
 
-    try {
-      final hr = ptr.ref.lpVtbl.value
-          .elementAt(6)
-          .cast<
-              Pointer<
-                  NativeFunction<
-                      HRESULT Function(
-            Pointer,
-            Pointer<Pointer<COMObject>>,
-          )>>>()
-          .value
-          .asFunction<
-              int Function(
-            Pointer,
-            Pointer<Pointer<COMObject>>,
-          )>()(
-        ptr.ref.lpVtbl,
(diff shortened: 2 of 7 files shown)


timsneath:
Wow! This is awesome. I'm on vacation and looking on my phone, but can't wait to review this properly in a couple of days.

timsneath:
Added a few tests and tried to simplify. I'm still trying to unpack the layers here to understand why this is a special case, but it certainly fixes what's broken right now.

timsneath:
I think I might merge this in its current state, since it's an improvement and then refactor further. How do you feel  about that? Are you happy with the tweaks I've made?

halildurmus:
> I think I might merge this in its current state, since it's an improvement and then refactor further. How do you feel  about that? Are you happy with the tweaks I've made? 

Thanks Tim, it sounds good.

PR #259, opened 2021-06-19
Support `Windows.Foundation.Collections`

Sunbreak:

PR #406, opened 2022-04-15
Manually project `IVector` and `IVectorView`

halildurmus:
Currently, the only supported types are `String` and `WinRT` types such as `IHostName`. 

I found the types that `IVector` and `IVectorView` support in [here](https://docs.rs/winrt/0.6.0/src/winrt/rt/gen/windows/foundation.rs.html#4986). 

From the link above, the types that `IVector` and `IVectorView` support are `WinRT` types, primitive types like `String`, `int`, `bool` etc., and also some structs like `GUID`, `Point`, `Size`, `DateTime`, `Uri` etc. 

I looked into the `WinMD` files using `ILSpy` and the majority of the APIs returns `IVector<String>` or `IVector<T extends IInspectable>`. This is also the case for `IVectorView`. That's why I didn't add support for the remaining types and some of them were structs like `GUID`, `Point` which I don't know how to project them. We can add support for them in the future.

While adding support for generic types, I had to find a way to create constructors from generic `T` types. In the beginning, I was storing the constructor tear-off mappings for `T` types inside a `Map` like this:

```dart
  static const map = <Type, IInspectable Function(Pointer<COMObject>)>{
    IHostName: IHostName.new,
    IStorageFile: IStorageFile.new,
    ...
  };
```
After learning the number of types that `IVectorView` supports -- which is more than `300`, I abandoned this plan and added a `creator` parameter to constructors of `IVector` and `IVectorView`.  

This parameter is only required for `WinRT` types like `IVector<IHostName>` and as I assume we will convert methods or properties that return `Pointer<COMObject>` to their respective types such as `IVector` or `IVectorView`(we're gonna expose this as a `List`, but first we're gonna turn this into an `IVectorView` in the generated code), the consumers won't have to deal with this which is great.

I'm looking forward to hearing your thoughts on this.

**EDIT:** I just figured out how to use `IIterable` and `IIterator`. I won't include them in this since this PR has grown quite a bit already.


Sunbreak:
Good job. https://github.com/timsneath/win32/pull/259 depend on this!

timsneath:
Been playing around with this, with a couple of suggestions here: https://github.com/halildurmus/win32/pull/1

timsneath:
My goodness: this is impressive. You've taken this a whole level further. 

I'll review the PR, but it'll take me a little while given its length. 

Would you mind signing a CLA, by the way? https://cla.developers.google.com/about/google-individual

halildurmus:
> My goodness: this is impressive. You've taken this a whole level further.
> 
> I'll review the PR, but it'll take me a little while given its length.

Thanks.

> Would you mind signing a CLA, by the way? https://cla.developers.google.com/about/google-individual

I signed this last year.

lib/src/extensions/iinspectable_query.dart
```diff
@@ -0,0 +1,35 @@
...
+  /// Queries `IInspectable` with the given `iid` for the relevant interface.
+  ///
+  /// Throws a `WindowsException` if the interface is not found.
+  ///
+  /// It is the caller's responsibility to deallocate the returned pointer when
+  /// they are finished with it. A FFI `Arena` may be passed as a
+  /// custom allocator for ease of memory management.
+  ///
+  /// {@category winrt}
+  Pointer<COMObject> query(String iid, {Allocator allocator = calloc}) {
```
timsneath: `IUnknown` has a `toInterface()` method that does this already. It's also not specific to a WinRT (`IInspectable`) object, but generic to any COM object. Could you take a look at `IUnknown.dart` and see if that meets your needs? Feel free to adapt accordingly.

lib/src/extensions/comobject_pointer.dart
```diff
@@ -0,0 +1,44 @@
...
+  /// `length` must be equal to the number of elements stored inside the
+  /// `Pointer<COMObject>`.
+  ///
+  /// ```dart
+  /// ...
+  /// final list = pComObject.toListOf<IHostName>(IHostName.new, length: 4);
+  /// ```
+  ///
+  /// {@category winrt}
+  List<T> toListOf<T>(T Function(Pointer<COMObject>) creator,
```
timsneath: I don't think the `Of` here is idiomatic Dart? Should probably be `toList<T>`.

lib/src/extensions/hstring_array.dart
```diff
@@ -0,0 +1,37 @@
...
+  /// `length` must be equal to the number of elements stored inside the
+  /// `Pointer<HSTRING>`.
+  ///
+  /// ```dart
+  /// ...
+  /// final list = pHString.toList(length: 5);
+  /// ```
+  ///
+  /// {@category winrt}
+  List<String> toList({int length = 1}) {
```
timsneath, line 25: Given your (very nice) method in `comobject_pointer.dart`, I'd be tempted to merge this into that, either with an `@override` method or by testing the types and using this specialization if `Pointer<COMObject>` matches `Pointer<HSTRING>` and `T` is `String`.

lib/src/combase.dart
```diff
@@ -23,6 +23,11 @@ import 'utils.dart';
 class COMObject extends Struct {
   external Pointer<Pointer<IntPtr>> lpVtbl;
 
+  /// Sets the vtable's pointer to `ptr`.
+  void setLpVtbl(Pointer<Pointer<IntPtr>> ptr) {
```
timsneath: Curious: why do you need this? Why not simply set `lpVtbl` directly? As you probably know, in Dart, it's not idiomatic to wrap fields with properties in the way that a language like C# does.

lib/win32.dart
```diff
@@ -263,6 +268,7 @@ export 'src/com/ITypeInfo.dart';
 export 'src/com/IUnknown.dart';
 export 'src/com/IUri.dart';
 export 'src/com/IUserDataPathsStatics.dart';
+export 'src/com/IVector`1.dart';
```
timsneath, line 254: Are you deliberately not exporting `IVectorView``1.dart` here?

test/winrt_collections_test.dart
```diff
@@ -0,0 +1,416 @@
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:test/test.dart';
+import 'package:win32/src/com/IVectorView%601.dart';
+import 'package:win32/win32.dart';
+
+// Exhaustively test the WinRT Collections to make sure constructors,
```
timsneath, line 7: Superb!

test/winrt_collections_test.dart
```diff
@@ -0,0 +1,416 @@
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:test/test.dart';
+import 'package:win32/src/com/IVectorView%601.dart';
```
timsneath: We shouldn't need this line if IVectorView is exported from win32.dart.

tool/inputs/interfaces.dart
```diff
@@ -103,6 +103,8 @@ const comInterfacesToGenerate = <String>[
 ];
 
 final windowsRuntimeTypesToGenerate = [
+  'Windows.Networking.Connectivity.INetworkInformationStatics',
```
timsneath: For documentation purposes, I'd like to include a commented out section in this set for manually-generated WinRT interfaces, just to be clear where they are coming from. Feel free to remove the other commented rows too -- they're just things I was experimenting with.

timsneath (changes requested):
I've made a few relatively minor comments; if you can address these, I think we'll be in good shape to merge and then iterate further.

tool/inputs/interfaces.dart
halildurmus: Done

test/winrt_collections_test.dart
halildurmus: Done

test/winrt_collections_test.dart
halildurmus: Thank you!

lib/win32.dart
halildurmus: I was. Since we were gonna expose methods or properties that return `IVectorView` as a `List` to consumers, I thought it would be a good idea to not export `IVectorView` to consumers. After your question, I remembered that some functions take `IVectorView` as parameters, so I exported it.

lib/src/combase.dart
halildurmus: Right. I forgot that we can set the `lpVtbl` parameter directly. Nice catch.

lib/src/extensions/comobject_pointer.dart
halildurmus: Done

lib/src/extensions/iinspectable_query.dart
halildurmus: Good catch. How did I miss that? :sweat_smile:

lib/src/extensions/hstring_array.dart
halildurmus: I'm confused. `Pointer<COMObject>` can't match `Pointer<HSTRING>` can it? They're very different types. 

`toList` extension method in `comobject_pointer.dart` exposed to `Pointer<COMObject>` types but not to `Pointer<HSTRING>` types. 

So I don't understand how can I override that method in `hstring_array.dart` or check that if `Pointer<COMObject>` matches to `Pointer<HSTRING>`.

halildurmus:
Hey @timsneath, this test is flaky: 
https://github.com/timsneath/win32/blob/cc9933f2216b5da247227a050aa3e2f21421abaf/test/bstr_test.dart#L23-L26

If you run this test several times on your computer, you will notice that it sometimes fails.

lib/src/extensions/hstring_array.dart
timsneath: No, I'm just being dumb. You're right. In a mad moment, I was thinking you could test if `COMObject` is `HSTRING`, but of course it never is 🙃 Ignore this comment.

lib/win32.dart
timsneath: Yeah, I think it's good to expose it even if we wind up completely wrapping it for convenience.

timsneath:
Are we good to merge, do you think? (Aside from my own test flake...)

halildurmus:
Yes, I think we're good.

timsneath (approved):

timsneath:
Thank you for this amazing pull request, Halil! This is awesome 🥳

PR #418, opened 2022-05-05
Project `DateTime` get properties

halildurmus:


tool/inputs/interfaces.dart
```diff
@@ -112,13 +112,13 @@ final windowsRuntimeTypesToGenerate = [
   // changes to the generation script.
   // 'Windows.Foundation.IAsyncAction',
   // 'Windows.Foundation.IAsyncOperation`1',
+  // 'Windows.Globalization.ICalendar',
```
halildurmus, line 115: `ICalendar` should not be autogenerated until we update the generation script to expose `IVectorView` types as `List`.

tool/inputs/interfaces.dart
```diff
@@ -112,13 +112,13 @@ final windowsRuntimeTypesToGenerate = [
...
   // 'Windows.Networking.Connectivity.INetworkInformationStatics',
-  // 'Windows.Networking.IHostName',
   // 'Windows.Storage.IUserDataPathsStatics',
   // 'Windows.Storage.Pickers.IFileOpenPicker',
   //
   'Windows.Foundation.IAsyncInfo',
   'Windows.Foundation.IClosable',
   'Windows.Foundation.IStringable',
-  'Windows.Globalization.ICalendar',
+  'Windows.Networking.IHostName',
```
halildurmus, line 123: `IHostName` can be generated automatically.

timsneath:
Awesome! At some point, we'll need to add support for parameters and setters. The former could be a little more complex, since of course a method could include a DateTime, a HSTRING, and a Vector...

halildurmus:
> Awesome!

Thanks.

> At some point, we'll need to add support for parameters and setters. The former could be a little more complex, since of course a method could include a DateTime, a HSTRING, and a Vector...

Yeah, adding support for parameters is gonna be really tricky. I'm currently working on projecting `IMap` and `IMapView` types. Once I'm done with that I might take a stab at it.

Issue #417, opened 2022-05-04
Inheritance is broken for `WinRT` types


halildurmus:
I noticed this when trying to make async calls to `WinRT`. I believe this affects all `WinRT` types that extend a type other than `IInspectable`. e.g. `IAsyncOperation`(extends IAsyncInfo), `IStorageFile`(extends IStorageItem).

As of now, the generation script calculates `vtable` start index by iterating all of the interfaces that an interface extends and summing the number of methods in those interfaces: 
https://github.com/timsneath/win32/blob/dcec79e0d1d0de9fc1a8c1836ad1cd96ff7c9b72/tool/projection/com_interface.dart#L32-L34

With this calculation, `IAsyncOperation`'s vtable start index becomes `11` which I think it's wrong and I believe it should be `6`. The function should only iterate for `IInspectable` and `IUnknown` interfaces. 

The reason for that is I was looking into the header(.h) file for `Windows.Foundation` [here](https://raw.githubusercontent.com/tpn/winsdk-10/master/Include/10.0.16299.0/winrt/windows.foundation.h) and I found `IAsyncOperation`'s vtable:
```c
typedef struct __FIAsyncOperation_1_IInspectableVtbl
{
    BEGIN_INTERFACE
    HRESULT ( STDMETHODCALLTYPE *QueryInterface )(__RPC__in __FIAsyncOperation_1_IInspectable * This,
        /* [in] */ __RPC__in REFIID riid,
        /* [annotation][iid_is][out] */ 
        _COM_Outptr_  void **ppvObject);
    ULONG ( STDMETHODCALLTYPE *AddRef )(__RPC__in __FIAsyncOperation_1_IInspectable * This);
    ULONG ( STDMETHODCALLTYPE *Release )(__RPC__in __FIAsyncOperation_1_IInspectable * This);

    HRESULT ( STDMETHODCALLTYPE *GetIids )(__RPC__in __FIAsyncOperation_1_IInspectable * This,
        /* [out] */ __RPC__out ULONG *iidCount,
        /* [size_is][size_is][out] */ __RPC__deref_out_ecount_full_opt(*iidCount) IID **iids);
    HRESULT ( STDMETHODCALLTYPE *GetRuntimeClassName )(__RPC__in __FIAsyncOperation_1_IInspectable * This, /* [out] */ __RPC__deref_out_opt HSTRING *className);
[...]

PR #423, opened 2022-05-10
Implement get `iids` property

halildurmus:
Implemented get `iids` property.

### DIFF
--- a/lib/src/winrt_helpers.dart
+++ b/lib/src/winrt_helpers.dart
@@ -176,9 +176,27 @@ enum TrustLevel {
 }
 
 extension IInspectableExtension on IInspectable {
-  // TODO: Add iids property when IVectorView`1 work lands.
+  /// Returns the interface IIDs that are implemented by the current Windows
+  /// Runtime class.
+  ///
+  /// The `IUnknown` and `IInspectable` interfaces are excluded.
   List<String> get iids {
-    throw UnimplementedError();
+    final pIIDCount = calloc<Uint32>();
+    final pIIDs = calloc<Pointer<GUID>>();
+
+    try {
+      final hr = GetIids(pIIDCount, pIIDs);
+      if (SUCCEEDED(hr)) {
+        return [
+          for (var i = 0; i < pIIDCount.value; i++) pIIDs.value[i].toString()
+        ];
+      } else {
+        throw WindowsException(hr);
+      }
+    } finally {
+      free(pIIDCount);
+      free(pIIDs);
+    }
   }
 
   /// Gets the fully qualified name of the current Windows Runtime object.
--- a/test/winrt_test.dart
+++ b/test/winrt_test.dart
@@ -51,6 +51,26 @@ void main() {
       winrtUninitialize();
     });
 
+    test('WinRT getIids test', () {
+      winrtInitialize();
+
+      const calendarClassName = 'Windows.Globalization.Calendar';
+      const iids = [
+        '{CA30221D-86D9-40FB-A26B-D44EB7CF08EA}', // ICalendar
+        '{00000038-0000-0000-C000-000000000046}', // IWeakReferenceSource
+        '{BB3C25E5-46CF-4317-A3F5-02621AD54478}', // ITimeZoneOnCalendar
+        '{0CA51CC6-17CF-4642-B08E-473DCC3CA3EF}'
+      ];
+
+      final object = CreateObject(calendarClassName, IID_ICalendar);
+      final calendar = ICalendar(object);
+
+      expect(calendar.iids, equals(iids));
+
+      free(object);
+      winrtUninitialize();
+    });
+
     test('WinRT getRuntimeClassName test', () {
       winrtInitialize();
 


timsneath:
Curious -- how do you see this being used? (I'm wondering if this should be hung off every WinRT class as a public field, or whether it's a helper that's only useful during the projection process?)

halildurmus:
> Curious -- how do you see this being used?

I can see this being used mainly during the projection process. I think this is SUPER helpful.

> (I'm wondering if this should be hung off every WinRT class as a public field...

If you're suggesting exposing `iids` as a `List` by autogenerating them in the projection script like this:
```dart
class IFileOpenPicker extends IInspectable {
  static const iids = [
    '{00000038-0000-0000-C000-000000000046}', // IWeakReferenceSource
    '{2CA8278A-12C5-4C5F-8977-94547793C241}', // IFileOpenPicker
    '{8CEB6CD2-B446-46F7-B265-90F8E55AD650}', // IFileOpenPicker2
    '{3F57B569-2522-4CA5-AA73-A15509F1FCBF}', // IFileOpenPickerWithOperationId
    '{D9A5C5B3-C5DC-5B98-BD80-A8D0CA0584D8}', // IFileOpenPicker3
  ];

  ...
}
```

... we can't do that because interfaces require specific Windows SDK versions. For example `IFileOpenPicker3` was added on
Windows SDK Version `18362` and my computer's build number is `19043`. If I run the generation script, IFileOpenPicker class will include `IFileOpenPicker3`'s IID inside the `iids` list since my computer's build number is greater than `IFileOpenPicker3`'s requirement.  

Suppose there is a Windows computer with a build number lower than `18362`. That computer can't use the `IFileOpenPicker3` interface, but the `iids` list inside the class will list `IFileOpenPicker3`'s iid, which will mislead the consumer that the interface is supported.

I got bitten by this in this PR. In the beginning, I wrote a test for `IFileOpenPicker`'s iids property. But the test checks failed for `(windows-2019, main)`. After some investigation, I found out that the machine didn't produce a `List` containing `IFileOpenPicker3` because the machine's build number was `17763`(which is lower than the requirement for `IFileOpenPicker3`). I solved this by changing the test to check for `ICalendar`(which doesn't have a recently added interface)'s iids.

timsneath:
I see -- yeah, that makes sense about it being version dependent. 

I suppose I'm merely wondering whether it makes sense to expose it as a public property on all WinRT interfaces. Does it clutter up the user's IntelliSense, for example? I'm wondering whether it would be better as a private function or property that the projection uses (e.g. within a method that casts to other interfaces). I could see it being more useful to a user if it listed the interfaces themselves, rather than the GUIDs, perhaps? 

Thank you so much for the work you're doing here -- I've been a little busy over the last couple of weeks, but it's amazing to see how you're taking the WinRT work forward. It lets me focus on clearing up some of the other technical debt in the project! Let me know how it's helpful for me to collaborate.

halildurmus:
> I suppose I'm merely wondering whether it makes sense to expose it as a public property on all WinRT interfaces. 

You suggested using `mixins` to project interfaces that are implemented by a `WinRT` class in [here](https://github.com/timsneath/win32/issues/417#issuecomment-1119269117). I played with this a little bit and I think this is a great idea.

Take the `Calendar` class for example. By using mixins, we could project it like this:
```dart
class Calendar extends IInspectable with ICalendar, ITimeZoneOnCalendar {
  Calendar(Pointer<COMObject> ptr) : super(ptr);
}
```

Mixins like the `ITimeZoneOnCalendar` would be projected something like this:
```dart
mixin ITimeZoneOnCalendar on IInspectable {
  // Use this ptr within the properties and methods of this mixin.
  late final _ptr = toInterface(IID_ITimeZoneOnCalendar);

  // Properties and methods of this mixin.
  ...
}
```

By doing that, we would expose all of the interfaces that are implemented by a WinRT class to consumers so they wouldn't need to deal with the `iids` property and the `toInterface` helper function.

Of course, we would need to create a helper function for activating a `WinRT` class because we can't use the `CreateObject` helper function since that function queries the given interface's `iid` and returns that interface's pointer to the caller. We need the original pointer.

To activate a `WinRT` class, we can implement a function like this:
```dart
Pointer<COMObject> ActivateWinRTClass(String className,
    {Allocator allocator = calloc}) {
  final hstrClass = calloc<HSTRING>();
  final inspectablePtr = allocator<COMObject>();
  final classNamePtr = className.toNativeUtf16();

  try {
    // Create a HSTRING representing the object
    var hr = WindowsCreateString(classNamePtr, className.length, hstrClass);
    if (FAILED(hr)) {
      throw WindowsException(hr);
    }
    // Activates the specified Windows Runtime class. This returns the WinRT
    // IInspectable interface, which is a subclass of IUnknown.
    hr = RoActivateInstance(hstrClass.value, inspectablePtr.cast());
    if (FAILED(hr)) {
      throw WindowsException(hr);
    }

    // Return a pointer to the relevant class
    return inspectablePtr;
  } finally {
    free(classNamePtr);
    free(hstrClass);
  }
}
```

And to make it easier to use the activatable `WinRT` classes, we could also consider generating `factory` constructors for them:
```dart
class Calendar extends IInspectable with ICalendar, ITimeZoneOnCalendar {
  Calendar(Pointer<COMObject> ptr) : super(ptr);

  factory Calendar.create() =>
      Calendar(ActivateWinRTClass('Windows.Globalization.Calendar'));
}
```

Consumers would then use the factory constructors to create an instance of an activatable `WinRT` class:
```dart
  final calendar = Calendar.create();
  print(calendar.GetDateTime());

  free(calendar.ptr);
```

> Does it clutter up the user's IntelliSense, for example? I'm wondering whether it would be better as a private function or property that the projection uses (e.g. within a method that casts to other interfaces). I could see it being more useful to a user if it listed the interfaces themselves, rather than the GUIDs, perhaps?

Unfortunately, I think it does. For example, these are the methods and properties that are exposed by the `IInspectable`:



We could consider making `GetIids`, `GetRuntimeClassName`, and `GetTrustLevel` functions private and moving the extension properties to the `IInspectable` class. That would reduce the clutter within the user's IntelliSense.

> Thank you so much for the work you're doing here -- I've been a little busy over the last couple of weeks, but it's amazing to see how you're taking the WinRT work forward. It lets me focus on clearing up some of the other technical debt in the project!

Thank you for the kind words. I also want to thank you for your work on this package. You've done a great job with it so far and I'm happy to be able to contribute to it. Dart is my favorite programming language. All of this `FFI` and `WinRT` stuff is really exciting to me and I really enjoy working on this.

> Let me know how it's helpful for me to collaborate.

I'm currently working on making async calls to WinRT and the projection `IMap` and `IMapView`. There are some design decisions to be made regarding these. I'll need to hear your thoughts on them when I send out PRs in the coming weeks.

timsneath:
I love this. I wonder: as we're starting to it should be `Calendar()` rather than using the factory `Calendar.create()`. That feels more idiomatic. That will cause a cascading breaking change for traditional COM interfaces though, and I'm not sure how to get around that. Perhaps they need to become factory constructors themselves, with `IFileOpenDialog.cast(myComObject)` or something. We should probably move this dialog into a separate design doc 

In the meantime, I'm going to merge this, since it's obviously in the same vein as `trustLevel` and `runtimeClassName`. Maybe we'll hide these from the "outside world" later on. It looks like that's what C# does.

PR #438, opened 2022-05-23
Fix Calendar example

halildurmus:


timsneath:
Oops -- I missed that! Thank you for catching. LGTM!

PR #439, opened 2022-05-23
Use `createFromID` helper method in factory constructors

halildurmus:
Also converted the filenames in the header to lowercase.


timsneath:
Great! Thanks!

PR #441, opened 2022-05-23
Project WinRT structs

halildurmus:
Added support for generating WinRT structs automatically. For now, I've only projected the structs within the `Windows.Foundation` namespace. I specifically need `Point`, `Size`, and `Rect` structs to extend the supported types of `IVector` & `IVectorView` and also for adding support for the `IMap` & `IMapView` interfaces.

### DIFF
--- a/example/shell_notify_icon/_menu.dart
+++ b/example/shell_notify_icon/_menu.dart
@@ -2,7 +2,7 @@ import 'dart:ffi';
 import 'dart:math';
 
 import 'package:ffi/ffi.dart';
-import 'package:win32/win32.dart';
+import 'package:win32/win32.dart' hide Point;
 
 import '_app.dart' as app;
 import '_tray.dart' as tray;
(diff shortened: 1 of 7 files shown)


timsneath:
Merged, but I wonder if the name clashes are going to be problematic over time. This has not been much of a problem with Win32 because of the `SCREAMING_CAPS` convention, but obviously things like `Point` and `Size` are much more challenging. 

I wonder if WinRT is going to be better as a discrete library export so folk can do something like:

```dart
import 'package:win32/winrt.dart' as winrt;
```

Any thoughts?

halildurmus:
> Merged, but I wonder if the name clashes are going to be problematic over time. This has not been much of a problem with Win32 because of the `SCREAMING_CAPS` convention, but obviously things like `Point` and `Size` are much more challenging.

Yeah, I'm afraid it will as the WinRT projection matures over time. For example, there is the class [Uri](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.uri?view=winrt-22000) which will clash with `dart:core`'s `Uri` once we add support for it.

> I wonder if WinRT is going to be better as a discrete library export so folk can do something like:
> 
> ```dart
> import 'package:win32/winrt.dart' as winrt;
> ```
> 
> Any thoughts?

I think that's a great idea.

PR #442, opened 2022-05-25
Complete (manual) projection of Calendar

timsneath:


halildurmus:
Awesome! 

I think `IAsyncAction` and `IAsyncOperation` should stay as a `class` since asynchronous APIs return one of these interfaces directly.

https://github.com/timsneath/win32/blob/715844987207a1a66e2a6a36d1be4dd114d1d87a/test/winrt_calendar_test.dart#L457-L466
The reason this test doesn't work is that creating our own `IVector` is not yet supported (forgot to mention that in the PR). I've spent a lot of time on this but haven't figured it out yet.

Until we figure it out, you can use the vector returned from the `IFileOpenPicker`'s `FileTypeFilter` as a workaround.

This works:
```dart
    test('Calendar.CreateCalendarWithTimeZone constructor', () {
      final picker = FileOpenPicker(
          ActivateClass('Windows.Storage.Pickers.FileOpenPicker'));
      final languages = picker.FileTypeFilter..ReplaceAll(['en-US', 'en-GB']);
      // ignore: constant_identifier_names
      const IID_IIterable_String = '{E2FCC7C1-3BFC-5A0B-B2B0-72E769D1CB7E}';
      final pIterable = languages.toInterface(IID_IIterable_String);

      try {
        final customCal = Calendar.CreateCalendarWithTimeZone(pIterable,
            'GregorianCalendar', '24HourClock', 'America/Los_Angeles');

        expect(customCal.Languages, equals(['en-US', 'en-GB']));
        expect(customCal.GetCalendarSystem(), equals('GregorianCalendar'));
        expect(customCal.GetClock(), equals('24HourClock'));
        expect(customCal.GetTimeZone(), equals('America/Los_Angeles'));
      } finally {
        free(picker.ptr);
        free(languages.ptr);
        free(pIterable);
      }
    });
```

timsneath:
Yeah, that's the limitation of `mixins`, in that they can't be directly instantiated. Conversely, the lack of multiple inheritance means that `extends` doesn't work for our use case. I was working on the basis that `*Statics` and `*Factory` should be classes, but I'm not certain how else to reason about other interfaces.

halildurmus:
I've been thinking about this. There are so many APIs that return interfaces such as [StorageFolder.GetItemAsync()](https://docs.microsoft.com/en-us/uwp/api/windows.storage.storagefolder.getitemasync?view=winrt-22000) (returns `IAsyncOperation<IStorageItem>`) and I don't think we can continue with the `mixins` because of this.

I've been playing with the idea of using `implements` and I think it's working pretty well so far. I've manually projected `Calendar` with this design in [here](https://github.com/halildurmus/win32/blob/winrt-class/lib/src/winrt/calendar.dart). Modifying the generator to support this might be tricky though.

What do you think?

PR #445, opened 2022-05-27
Update WinRT generation

halildurmus:
- Create a default constructor for WinRT Class only if it supports it. (I do this by checking if the class has an `ActivatableAttribute` with `two` parameters. This is because if a class has a default constructor, it has this attribute: `[Activatable(65536u, "Windows.Foundation.UniversalApiContract")]`)
- Removed `guidConstants` as WinRT classes don't have any.
- Removed `_thisPtr` as it's redundant now.

Thank you for your work on the WinRT Class projection, it's great.


timsneath:
Great! Looks good, thank you. The `_thisPtr` work inspired me to create #446.

PR #454, opened 2022-05-30
Update class declaration to fix the test

halildurmus:


timsneath:
Sorry -- that was my bad. I fixed it with #455 already, I think?

halildurmus:
Yeah, it's okay :)

PR #465, opened 2022-05-31
Add basic support for WinRT asynchronous methods

### LINKED ISSUE #39: Add WinRT FileOpenPicker
Depends on #97

### LINKED ISSUE #97: Figure out async calls to WinRT
As an example, this doesn't work yet:
https://github.com/timsneath/win32/blob/main/example/winrt_picker.dart#L20

FileOpenPicker is here:
https://github.com/timsneath/win32/blob/main/lib/src/com/IFileOpenPicker.dart

which is derived using the Windows Runtime metadata using my Dart-based parser (https://pub.dev/packages/winmd). 

The original signature is here:
https://docs.microsoft.com/en-us/uwp/api/windows.storage.pickers.fileopenpicker.picksinglefileasync?view=winrt-19041

halildurmus:
Fixes #39 
Fixes #97 

Adds support for WinRT asynchronous methods that return [IAsyncAction](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncaction?view=winrt-22621) or [IAsyncOperation<TResult>](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncoperation-1?view=winrt-22621) delegates. This is achieved by polling their [status](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncinfo.status?view=winrt-22621) property every `10 milliseconds` (I don't have a strong opinion on what this value should be) until it completes.

The other WinRT projections use the [completed](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncoperation-1.completed?view=winrt-22621) callback but we can't use that at the moment as [async callbacks](https://github.com/dart-lang/sdk/issues/37022) are not supported yet. This also prevents us from adding support for the asynchronous methods that return [IAsyncActionWithProgress](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncactionwithprogress-1?view=winrt-22621) and [IAsyncOperationWithProgress](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncoperationwithprogress-2?view=winrt-22621) delegates (they report progress updates to callers through their [progress](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncoperationwithprogress-2.progress?view=winrt-22621) callback).

Furthermore, I added [IInitializeWithWindow](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-iinitializewithwindow) interface and a wrapper for it (in `winrt_com_interop_helpers.dart`). This is necessary for APIs that depend on `CoreWindow` ([Display WinRT UI objects that depend on CoreWindow](https://learn.microsoft.com/en-us/windows/apps/develop/ui-input/display-ui-objects) explains this).

Also added the `BasicProperties` class and partially generated the `StorageFile` APIs (I'll fully generate them later as StorageFile has so many dependencies).


timsneath:
Woah! Is this ready for review? That would be the _best_ Christmas present!

halildurmus:
> Woah! Is this ready for review? That would be the _best_ Christmas present!

Yeah, I think this is finally ready for review :).

example/winrt_picker.dart
```diff
@@ -2,74 +2,38 @@
...
-
-import 'package:ffi/ffi.dart';
 import 'package:win32/winrt.dart';
 
 void main() async {
   winrtInitialize();
-  final pIndex = calloc<Uint32>();
 
   final picker = FileOpenPicker()
     ..suggestedStartLocation = PickerLocationId.desktop
```
timsneath, line 13: I'm not sure why, but neither the `suggestedStartLocation` nor the `viewMode` property appear to be working on my machine. I've tried changing them to various other values, but they don't seem to affect the presented dialog box. Can you replicate?

lib/src/winrt/internal/async_helpers.dart
```diff
@@ -0,0 +1,63 @@
...
+      case AsyncStatus.started:
+        break;
+      case AsyncStatus.completed:
+        onCompleted();
+        break;
+      case AsyncStatus.canceled:
+        completer.completeError('The async operation canceled!');
+        break;
+      case AsyncStatus.error:
+        completer.completeError(WindowsException(asyncDelegate.errorCode));
```
timsneath, line 56: I think asyncDelegate.errorCode is an `HSTRING`, not an `HRESULT`, no? Based on this: https://learn.microsoft.com/en-us/uwp/api/windows.foundation.iasyncinfo.errorcode?view=winrt-22621#windows-foundation-iasyncinfo-errorcode

WindowsException takes an `HRESULT` and uses `FormatMessage` to convert it to a String.

lib/src/winrt_com_interop_helpers.dart
```diff
@@ -0,0 +1,49 @@
...
+/// {@category Class}
+/// {@category winrt}
+class InitializeWithWindow {
+  /// Specifies an owner window to be used by a Windows Runtime object that is
+  /// used in a desktop app.
+  ///
+  /// [target] must be a WinRT object that implements the
+  /// `IInitializeWithWindow` interface (e.g. `FileOpenPicker`).
+  ///
+  /// [hwnd] represents the handle of the window to be used as the owner window.
```
timsneath, line 26: I'm curious if there are only specific types of `HWND` that are valid. I used `GetActiveWindow()` (which I would have thought would be the natural choice), and got an error: `Error 0x80070578: Invalid window handle.`, which suggests that there's some specific criteria beyond being a valid `HWND`. I wasn't able to find documentation that answered this beyond [this article](https://learn.microsoft.com/en-us/windows/apps/develop/ui-input/display-ui-objects).

lib/src/winrt/internal/async_helpers.dart
```diff
@@ -0,0 +1,63 @@
...
+/// Completes the given [completer] by polling [asyncDelegate]'s `status`
+/// property every `10` milliseconds until the [asyncDelegate] completes.
+///
+/// [onCompleted] is called when the [asyncDelegate] completes successfully.
+Future<void> _completeAsyncDelegate<T extends IAsyncInfo, C>(T asyncDelegate,
+    Completer<C> completer, void Function() onCompleted) async {
+  try {
+    while (asyncDelegate.status == AsyncStatus.started) {
+      // Yield execution to Dart for 10 milliseconds to avoid blocking the app.
+      await Future<void>.delayed(const Duration(milliseconds: 10));
```
timsneath: As an interim measure until there's a better solution, I think I'd suggest 1/60s as the poll frequency, to match the default frame rate of 60fps. More regular than that, and it's unlikely that the user will see a result for UI-centric content, and from a battery perspective, we know that the processor will be active at that frequency. Something like: `const Duration(microseconds: 1*1000*1000/60.0)`, perhaps?

(If we knew this was being called from Flutter, we might use `addPersistentFrameCallback`, but of course we don't have that guarantee.)

timsneath (commented):
Added some general comments above. Hopefully helpful!

example/winrt_picker.dart
halildurmus: [FileOpenPicker.SuggestedStartLocation Property](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.fileopenpicker.suggestedstartlocation?view=winrt-22621#remarks) states:
> The SuggestedStartLocation is not always used as the start location for the file picker. To give the user a sense of consistency, the file picker remembers the last location that the user navigated to and will generally start at that location.

As for the `viewMode` property, I can't get it to work either. I'm installing Visual Studio right now. I'll try this example in a UWP app to see if this works properly there.

lib/src/winrt/internal/async_helpers.dart
halildurmus: I'm pretty sure this returns an `HRESULT`. C# wraps this with `System.Exception`. 

If you change the language setting in the top right corner to `C++/WinRT`, you will see that it returns an HRESULT:
image

I also checked the `asyncinfo.h` file from the Windows SDK:
```h
 virtual /* [propget] */ HRESULT STDMETHODCALLTYPE get_ErrorCode( 
            /* [retval][out] */ __RPC__out HRESULT *errorCode) = 0;
```

lib/src/winrt_com_interop_helpers.dart
halildurmus: > I'm curious if there are only specific types of HWND that are valid. I used GetActiveWindow() (which I would have thought would be the natural choice), and got an error: Error 0x80070578: Invalid window handle., which suggests that there's some specific criteria beyond being a valid HWND. I wasn't able to find documentation that answered this beyond [this article](https://learn.microsoft.com/en-us/windows/apps/develop/ui-input/display-ui-objects).

I don't know. I found `GetConsoleWindow()` and `GetShellWindow()` by trial and error.

I just noticed that `GetForegroundWindow()` works for both console and Flutter apps (_yay!_). I wonder, should I make the necessary changes so that we use `GetForegroundWindow()` everywhere?

lib/src/winrt/internal/async_helpers.dart
halildurmus: Great, thanks!

example/winrt_picker.dart
halildurmus: Interesting, it doesn't work in a UWP app either.

lib/src/winrt/internal/async_helpers.dart
timsneath: Doh! You're totally right. It makes sense that the Microsoft language projections wrap various low-level concepts, but it's so confusing that their docs then tell lies about the interface signatures! I want a drop-down menu choice for "tell me the truth, not some language-specific warped reality" :)

lib/src/winrt_com_interop_helpers.dart
timsneath: `GetForegroundWindow()` is good, but it's not perfect -- for example, if there's another window from another process that has gained focus, it will become the parent for the file picker. That's why I'm a bit bemused that `GetActiveWindow()` isn't working here, since that means "the current window from the current process". (More here: https://devblogs.microsoft.com/oldnewthing/20081006-00/?p=20643)

halildurmus: Ah, I see. It was too good to be true anyway :).

example/winrt_picker.dart
timsneath: Very strange. Investigating here too...

timsneath (approved):
There's probably a little more iteration to be done on the example, but I think we're good to merge this as a baseline. This is awesome work, @halildurmus. I don't think I could have gotten there myself. Thank you so much!

halildurmus:
> There's probably a little more iteration to be done on the example, but I think we're good to merge this as a baseline. This is awesome work, @halildurmus. I don't think I could have gotten there myself. Thank you so much!

Thank you. Let's do this 🎉.

lib/src/winrt_com_interop_helpers.dart
timsneath: This seems to explain it:
https://devblogs.microsoft.com/oldnewthing/20190412-00/?p=102413

(TL;DR Not every window has a `CoreWindow` associated with it. We ought to figure out how to do `IInitializeWithWindow` from Dart).

clarkezone:
Awesome would love to give this a try!

Issue #463, opened 2022-05-31
Seeing occasional flakes during `dart test` on GitHub Actions


timsneath:
It's rare, but sometimes a test fails on a particular (platform,branch) combination with an error. 

For example: https://github.com/timsneath/win32/runs/6676600994?check_suite_focus=true

failed with:
```
Error: Process completed with exit code 1.
```

It's possible that there's a bug in the test suite (over-enthusiastic `free`, perhaps?) Or worse, that there's a bug in the COM wrappers themselves...

halildurmus:
Yeah, I've been observing this behavior for a while. It seems that it always happens after this test: https://github.com/timsneath/win32/blob/f63ecce439a8401d1bb326601c57584d77aa6c9d/test/com_network_test.dart#L28-L32

Since it's rare, it's really hard to replicate. I might look into it tomorrow.

timsneath:
I've been playing with the tests, adding multiple test runs in various cases, but can't get it to reproduce locally. It only seems to occur on the CI infra. Are you able to repro it locally, @halildurmus?

halildurmus:
Sadly, I couldn't reproduce it locally.

timsneath:
Closing this one for now in the hope that we've fixed it with the `NativeFinalizer` work. Please reopen if you observe this with builds later than v4.1.0.

PR #467, opened 2022-06-01
Fix WinRT structs projection

### LINKED ISSUE #466: Running `tools\generate.cmd` deletes the WinRT content in structs.g.dart
Something went wrong here when I added the structs used by `Gamepad`. Need to figure out what's going on here.

halildurmus:
Closes #466 

There might be room for improvement in this function but for now, this should fix the issue.


halildurmus:
#463 again :(

PR #479, opened 2022-06-03
Generate `implementsMappers` for WinRT Interfaces

### LINKED ISSUE #473: Project `implementsMappers` for WinRT Interfaces
While working on #464 today, I realized we should also project `implementsMappers` for WinRT Interfaces as they require other interfaces too (e.g. `IStorageFolder` requires `IStorageItem`).

halildurmus:
Closes #473

Also, I regenerated `ApplicationData` class and `IApplicationData` interface as they were a bit outdated.

### DIFF
--- a/example/storage.dart
+++ b/example/storage.dart
@@ -1,16 +1,15 @@
 import 'package:win32/win32.dart';
 
 void main() {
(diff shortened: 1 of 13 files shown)


timsneath:
Nice!

PR #484, opened 2022-06-06
Tidy up unnecessary version specific code

timsneath:
The generator contains some special exception code for when it's used to generate the _full_ Win32 API, rather than just the hand-selected APIs and structs that we define now. Working to remove this specific code and/or make it generic (usable in either scenario). 

These changes should make no difference to the results of tool\generate.cmd.


halildurmus (approved):
Nice!

PR #487, opened 2022-06-06
Fix `generateWinRTApis()`

halildurmus:
In #476, I introduced a bug that prevents all generic interfaces from being generated. I only intended to block the generation of generic interfaces from the type's dependencies (if any), not the type itself. This fixes it.


timsneath (commented):
Might be worth splitting this up into a few statements, with a comment to describe each, rather than one long chain. Would be easy to regress here, with all the nested braces and brackets.

halildurmus:
Yeah, you're right. I'll do that.

timsneath (approved):
Oh, this is much cleaner! Thank you!

PR #490, opened 2022-06-07
Update pubspec to support reftypes in winmd

timsneath:
This version of winmd should be neutral to the generator, but will solve the problem where reference types weren't projected properly.


halildurmus (approved):
Great, thank you!

PR #491, opened 2022-06-07
Clean up `specialTypes`

halildurmus:
I've added these before with #405 and #406. It was necessary at the time because the `isInterface` property was returning false for these generic interfaces which were preventing the projection of these types: https://github.com/timsneath/win32/blob/05df9384ad9c4c51a59d3f3d4701ac38494a9494/tool/generator/lib/src/projection/type.dart#L275-L285

I don't know when this was fixed but it seems to be working properly now. Therefore, we can remove them from the `specialTypes` map.


timsneath (approved):
LGTM!

PR #492, opened 2022-06-08
Fix projection of reference types

halildurmus:
Depends on [this](https://github.com/timsneath/winmd/pull/58) PR.

Since reference types were broken before #490, we were projecting all reference types as `Pointer<COMObject>` which is incorrect for some types (e.g. `value` parameters of `CreateUint8Array` and `GetUint8Array`). Now that it's fixed, we can start projecting them correctly.


timsneath:
I think we're missing a pointer reference here. For example, `IPropertyValue.GetUint8Array()` should have parameters Pointer<Uint32>, Pointer<Pointer<Uint8>>. 

This works:
```dart
      final array = calloc<Uint8>(5);
      for (var idx = 0; idx < 5; idx++) {
        array[idx] = (10 * idx) + 10;
      }
      final pv = IPropertyValue.from(PropertyValue.CreateUInt8Array(5, array));
      final arraySize = calloc<Uint32>();
      final newArray = calloc<Pointer<Uint8>>();

      pv.GetUInt8Array(arraySize, newArray.cast());
      expect(arraySize.value, equals(5));
      expect(newArray.value[0], equals(10));
      expect(newArray.value[1], equals(20));
      expect(newArray.value[2], equals(30));
      expect(newArray.value[3], equals(40));
      expect(newArray.value[4], equals(50));
```

but note the extra cast necessary to coerce a `Pointer<Pointer<Uint8>>` into a `Pointer<Uint8>`.

halildurmus:
Yeah, you're right. It's also mentioned in [here](https://docs.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system#array-parameters):
> ReceiveArray. This style is used when the caller receives an array that was allocated by the method. In this style, the array size parameter and the array parameter are both out parameters. Additionally, the array parameter is passed by reference (that is, ArrayType**, rather than ArrayType*).

As the array parameter is passed by reference, we either need to use `GetXXArray` methods as you did in your example or we need to expose these parameters as `Pointer<Pointer<...>>`. 

I created [this](https://github.com/timsneath/winmd/pull/59) PR for exposing these types as `Pointer<Pointer<...>>` and updated this PR accordingly.

timsneath:
I _think_ we want to address this last bit in win32, rather than in winmd, no? It seems like winmd is projecting this correctly as a reference type, but it's win32 that needs to interpret this appropriately for the Dart projection. If it's OK with you, I'll try and update the PR with how I think it should be handled in win32, and then we can compare approaches?

halildurmus:
Sounds good 👍

timsneath:
The GUID test seems flaky. I haven't taken a proper look.

halildurmus:
I think it's because we're freeing the `retValuePtr` in the `finally` block of the `GetGuid()` function.

We probably need to create `structDeclaration()` method in `winrt_method.dart` for methods that return structs and don't free `retValuePtr` in it just like we do in `comObjectDeclaration()`.

EDIT: I skipped this test for now, we can fix this in a follow-up PR.

tool/generator/lib/src/projection/type.dart
```diff
@@ -251,6 +251,10 @@ class TypeProjection {
           'Pointer<${refTuple.nativeType}>', 'Pointer<${refTuple.dartType}>');
     }
 
+    if (typeIdentifier.typeArg?.baseType == BaseType.uint32Type) {
+      return const TypeTuple('Pointer<Uint32>', 'Pointer<Uint32>');
```
halildurmus, line 254: This handles parameters like [IndexOf(T, UInt32)](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.collections.ivectorview-1.indexof?view=winrt-22621#windows-foundation-collections-ivectorview-1-indexof(-0-system-uint32@))'s `index`:
```cs
  [MethodImpl(MethodImplOptions.InternalCall)]
  bool IndexOf([In] T value, out uint index);
```

We project these manually in IVector & IVectorView, but there may be other methods with reference Uint32 parameters that we haven't projected yet.

tool/generator/lib/src/inputs/interfaces.dart
```diff
@@ -107,6 +107,7 @@ final windowsRuntimeTypesToGenerate = <String>{
   'Windows.Data.Xml.Dom.IXmlNodeList',
   'Windows.Foundation.IAsyncInfo',
   'Windows.Foundation.IClosable',
+  'Windows.Foundation.IPropertyValue',
```
halildurmus, line 110: IPropertyValue is an outlier as PropertyValue doesn't depend on it (hence, it is not generated automatically), so we need to add this manually.

tool/generator/lib/src/projection/type.dart
```diff
@@ -224,6 +224,37 @@ class TypeProjection {
...
+  }
+
+  TypeTuple unwrapReferenceType() {
+    if ((typeIdentifier.typeArg?.type?.isInterface ?? false) ||
+        typeIdentifier.typeArg?.baseType == BaseType.classTypeModifier) {
+      return const TypeTuple('Pointer<COMObject>', 'Pointer<COMObject>');
+    }
+
+    if (typeIdentifier.typeArg?.baseType == BaseType.simpleArrayType) {
+      // TODO: Handle PassArray and FillArray
```
halildurmus: I think you can remove this TODO since the parameters that use these styles are exposed as simple array types, therefore, they are handled in the `unwrapSimpleArrayType()` method:
https://github.com/timsneath/win32/blob/bb73f777c4e4712bceb5c72ed3e2ea903484d3ad/tool/generator/lib/src/projection/type.dart#L302

PassArray style is used in PropertyValue's `CreateXXArray` methods and currently, they are projected correctly. We have several tests that validate this.

As for the FillArray style, it may not be possible to write tests for them as I couldn't find a method that uses this style other than `GetMany()` method from the IVectorView. Since IVectorView is a generic interface and GetMany() method takes a generic array as a parameter, the generator fails to project that method and throws an exception (type information missing for classVariableTypeModifier). 

I wanted to manually test to see if the GetMany() method would be generated correctly for String arrays if I returned `TypeTuple('IntPtr', 'int')` for classVariableTypeModifier in `type.dart` like this:
```dart
    ...
    if (isInterface ||
        typeIdentifier.baseType == BaseType.classTypeModifier ||
        typeIdentifier.baseType == BaseType.objectType) {
      return const TypeTuple('Pointer<COMObject>', 'Pointer<COMObject>');
    }

    if (typeIdentifier.baseType == BaseType.classVariableTypeModifier) {
      return const TypeTuple('IntPtr', 'int');
    }
    ...
```

Here is the generated method:
```dart
int GetMany(int startIndex, int valueSize, Pointer<IntPtr> value) {
    final retValuePtr = calloc<Uint32>();

    try {
      final hr = ptr.ref.vtable
              .elementAt(9)
              .cast<
                  Pointer<
                      NativeFunction<
                          HRESULT Function(
                              Pointer,
                              Uint32 startIndex,
                              Uint32 valueSize,
                              Pointer<IntPtr> value,
                              Pointer<Uint32>)>>>()
              .value
              .asFunction<
                  int Function(Pointer, int startIndex, int valueSize,
                      Pointer<IntPtr> value, Pointer<Uint32>)>()(
          ptr.ref.lpVtbl, startIndex, valueSize, value, retValuePtr);

      final retValue = retValuePtr.value;
      return retValue;
    } finally {
      free(retValuePtr);
    }
  }
```

The generated method is correct as the value parameter is wrapped with only `one` Pointer. You can see here the methods that I've generated manually in IVectorView to support retrieving COMObject and String arrays:
https://github.com/timsneath/win32/blob/fd55ddc5e1a1027a62eed567d0b229d4fab80235/lib/src/winrt/ivectorview.dart#L268-L348

tool/generator/lib/src/inputs/interfaces.dart
timsneath: That's most odd, isn't it. I see that now. `PropertyValue` is a static class. Thanks.

tool/generator/lib/src/projection/type.dart
timsneath: Thanks. I've added a comment to that effect.

tool/generator/lib/src/projection/type.dart
timsneath: Gotcha. Removed.

timsneath (approved):
Great! This shaped up really well!

PR #500, opened 2022-06-12
Remove overrides of operators `[]` and `[]=`

halildurmus:
I previously added these operators because I thought they weren't exposed for `Pointer<T extends Struct>` (e.g. `Pointer<COMObject>`) types. Looks like, I was wrong :) as they are defined [here](https://github.com/dart-lang/sdk/blob/5263e15e99d967bd9cec5f8b05fac95224bd768e/sdk/lib/ffi/ffi.dart#L682-L697).


timsneath:
Thanks. I'm away from my computer this weekend but I'll look this evening and tomorrow at the outstanding requests.

timsneath (approved):

PR #503, opened 2022-06-13
Fix projection of static setters

halildurmus:
Static setters should not return anything since their return types are void.

For example, prior to this change, Geolocator's DefaultGeoposition setter was incorrectly generated as:
```dart
  static set DefaultGeoposition(Pointer<COMObject> value) {
    final activationFactory =
        CreateActivationFactory(_className, IID_IGeolocatorStatics2);

    try {
      final result = IGeolocatorStatics2.from(activationFactory).DefaultGeoposition = value;
      return result;
    } finally {
      free(activationFactory);
    }
  }
```

With this change, it will be generated as:
```dart
  static set DefaultGeoposition(Pointer<COMObject> value) {
    final activationFactory =
        CreateActivationFactory(_className, IID_IGeolocatorStatics2);

    try {
      IGeolocatorStatics2.from(activationFactory).DefaultGeoposition = value;
    } finally {
      free(activationFactory);
    }
  }
```

### DIFF
--- a/lib/src/winrt/gamepad.dart
+++ b/lib/src/winrt/gamepad.dart
@@ -53,9 +53,7 @@ class Gamepad extends IInspectable
         CreateActivationFactory(_className, IID_IGamepadStatics);
 
     try {
-      final result =
-          IGamepadStatics.from(activationFactory).add_GamepadAdded(value);
-      return result;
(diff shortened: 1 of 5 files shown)


timsneath (approved):
LGTM!

PR #508, opened 2022-06-13
Exploring more idiomatic COM interface projections

### LINKED ISSUE #446: Cleaner COM instantiation and casting
We have the opportunity to reduce the ceremony with casting COM objects. 

Today, the code is something like:

```
  final fileDialog2 = IFileDialog2(
      COMObject.createFromID(CLSID_FileOpenDialog, IID_IFileDialog2));
  final modalWindow = IModalWindow(fileDialog2.toInterface(IID_IModalWindow));
```

This could be considerably simpler:

- Interfaces could have a `createInstance()` helper method in a same way to `FileOpenDialog.createInstance()`
- the queryInterface could take place automatically on a cast, such as `final modalWindow = IModalWindow.from(fileDialog2);`

(Inspired by the `_thisPtr` work that we experimented with on WinRT.)

timsneath:
This branch provides a `.from()` method for each interface that makes it easier to cast from one interface to another without dealing with IIDs.

It doesn't break WinRT, but the semantics of `.from()` are somewhat different. Need to resolve this before merging.


timsneath:
I'm struggling to reconcile this with the WinRT work. I reread #481, and I'm a bit confused, so I'd love to get your advice, @halildurmus!

This PR enables any COM (or WinRT) object to cast to another interface, for example:

```dart
  // fileOpenDialog is an instance of IFileOpenDialog; the COM object also implements IModalWindow.
  final modalWindow = IModalWindow.from(fileOpenDialog);
```

The `.from()` method here takes a single parameter of `IUnknown` (i.e. all COM and WinRT objects), and under the covers calls `QueryInterface` to find the relevant interface.

Ideally we'd gradually eradicate the need for package users to use `Pointer<COMObject>` directly: that would be masked away, and wherever that appears, we'd replace it with `IUnknown`, which is a container for the former. In doing so, we'd be able to have `IUnknown` implement `Finalizer` so that the underlying pointer could automatically be garbage collected. 

Conversely, your WinRT named constructor takes a `Pointer<COMObject>` that it presumes to be already the correct interface. Should we rename that to `.fromRawPointer()`, do you think? My head is hurting from trying to think this through...

halildurmus:
> Ideally we'd gradually eradicate the need for package users to use `Pointer<COMObject>` directly: that would be masked away, and wherever that appears, we'd replace it with `IUnknown`, which is a container for the former.

I _think_ we'd use `IInspectable` for WinRT types right?

> In doing so, we'd be able to have `IUnknown` implement `Finalizer` so that the underlying pointer could automatically be garbage collected.

I'm so excited about this!

> Conversely, your WinRT named constructor takes a `Pointer<COMObject>` that it presumes to be already the correct interface. Should we rename that to `.fromRawPointer()`, do you think? My head is hurting from trying to think this through...

Yeah, I think that renaming it to `.fromRawPointer()` makes sense.

I made some changes in this branch (hope it's okay :)). 

First, I moved the `.from` factory constructors to the top because I think it makes more sense. Then, I renamed all WinRT named constructors to `.fromRawPointer()`. After that, I modified the WinRT generation script to generate the `.from` for WinRT interfaces. And finally, I modified the implements mappers to use the `.from`. 

For example, instead of generating this:
```dart
  // ICalendar methods
  late final _iCalendar = ICalendar.fromRawPointer(toInterface(IID_ICalendar));
  ...
``` 

Now, we generate this which is much cleaner:

```dart
  // ICalendar methods
  late final _iCalendar = ICalendar.from(this);
  ...
```

timsneath:
This looks good to me, @halildurmus -- do you want to approve and merge?

halildurmus:
Sure.

halildurmus (approved):

PR #515, opened 2022-06-21
SetupDiGetDeviceRegistryProperty implementation and add SPDRP constants

hpstuff:


lib/src/constants.dart
```diff
@@ -109,10 +109,7 @@ const CLSCTX_REMOTE_SERVER = 0x10;
 
 /// The combination of `CLSCTX_INPROC_SERVER`, `CLSCTX_INPROC_HANDLER`,
 /// `CLSCTX_LOCAL_SERVER`, and `CLSCTX_REMOTE_SERVER`.
-const CLSCTX_ALL = CLSCTX_INPROC_SERVER |
-    CLSCTX_INPROC_HANDLER |
-    CLSCTX_LOCAL_SERVER |
-    CLSCTX_REMOTE_SERVER;
+const CLSCTX_ALL = CLSCTX_INPROC_SERVER | CLSCTX_INPROC_HANDLER | CLSCTX_LOCAL_SERVER | CLSCTX_REMOTE_SERVER;
```
timsneath: For consistency, we wrap using `dart format` with lines set to the default of 80 characters.

lib/src/constants.dart
```diff
@@ -8866,3 +8847,115 @@ class AUDCLNT_STREAMOPTIONS {
   static const AUDCLNT_STREAMOPTIONS_MATCH_FORMAT = 0x2;
   static const AUDCLNT_STREAMOPTIONS_AMBISONICS = 0x4;
 }
+
+/// Device registry property codes
+/// {@category Enum}
+class SPDRP {
+  /// DeviceDesc (R/W)
+  static const SPDRP_DEVICEDESC = 0x00000000;
```
timsneath, line 8876: For consistency, use the documentation from https://docs.microsoft.com/en-us/windows/win32/api/setupapi/nf-setupapi-setupdigetdeviceregistrypropertyw for these fields, which is CC licensed. Wrap at 80 characters, and strip out any unnecessary specifics about backward compatibility to older versions prior to Windows 7. You may omit undocumented values such as `SPDRP_UNUSED1` to avoid unnecessary bloat. 

For example:
```dart
  /// The function retrieves a CM_POWER_DATA structure that contains the
  /// device's power management information.
  static const SPDRP_DEVICE_POWER_DATA = 0x0000001E;
```

Thanks! I know it's a pain, but it's worth it for the long run!

timsneath (changes requested):
Looks great, modulo a couple of changes. Thanks for your contribution!

lib/src/constants.dart
hpstuff: yeah I saw it but wasn't able to find documentation about code formatting or something

hpstuff:
it's should be fine now, let me know if there anything else that I have to do.

timsneath (approved):
Great! Thank you!

PR #516, opened 2022-06-23
Convert method and property names to `camelCase`

### LINKED ISSUE #482: Return types conflicts with property names in getters and setters
There will be type conflicts when we start exposing their types correctly.

For example, here are the `Headset` and `User` getters of the `Gamepad`:
```dart
...
  @override
  Headset get Headset => _iGameController.Headset;

  @override
  User get User => _iGameController.User;
...
```

And here is the `NotificationMirroring` setter of the `ToastNotification`:
```dart
...
set NotificationMirroring(NotificationMirroring value) =>
      _iToastNotification3.notificationMirroring = value;
...
```

We may need to convert all property and method names to `camelCase`.

halildurmus:
Fixes #482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 3.0.0
+
+- [BREAKING CHANGE] COM and Windows Runtime methods and properties are now
+  camelCased, not TitleCased. This is inconvenient, but it avoids a whole class
+  of name clashes and aligns COM and WinRT APIs more closely with Dart idioms.
+  As the projections get smarter with more helpers, we think this is the right
+  call for the future and worth a one-time tax to fix.
+- TODO: Add other changes.
+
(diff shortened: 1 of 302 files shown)


timsneath:
If we're going to do this for WinRT, we also need to do it for COM for consistency, since WinRT is based on COM. Otherwise we'll have eccentricities like `QueryInterface` being TitleCased while other methods are camelCased.

I've proposed some changes that unify this. It's a fairly big change, so I've bumped the version to v3.0.0 in anticipation of a (hopefully soon) release.

timsneath (approved):

timsneath:
I've approved your changes but will leave this unmerged so you can review mine :) Feel free to merge in if you're happy with the net results!

halildurmus:
Awesome, LGTM!

PR #517, opened 2022-06-25
update ffi >=2.0.0 <3.0.0

richard457:
update ffi >=2.0.0 <3.0.0


timsneath:
Hmmm: is there a reason for this? This package is compatible with both v1.x and v2.x.

timsneath:
Closing for now. Feel free to file an issue if there's something that is not working that makes you think this is necessary.

PR #523, opened 2022-07-06
Support `int` and `WinRTEnum` types in `IIterator`, `IVector` and `IVectorView`

halildurmus:
Here is the summary of the changes in this PR:
- Adds support for `int` and `WinRTEnum` types in `IIterator`, `IVector` and `IVectorView`
- IVectorView's `toList()` now return `immutable` List using `List.unmodifiable()` (I believe this is the correct behavior)
- Adds `IIterable` IIDs into the `winrt_constants.dart`
- Refactor `IIterable`, `IIterator`, `IVector` and `IVectorView` interfaces
- Adds `DeviceClass`, `DevicePickerDisplayStatusOptions` and `Placement` enumerations
- Adds `DevicePicker`, `DevicePickerFilter` and `Printing3DMultiplePropertyMaterial` APIs (needed these for testing `IVector<DeviceClass>` and `IVector<int>`)
- Minor documentation improvements

First, I'd like to explain where I got these `IIterable` IIDs and why I added them. I got them from [here](https://github.com/Elmue/WindowsRT-GUID-Analyzer/blob/main/Release/Windows%2011%20(Build%2022000)%20WinRT%20Interfaces.ini) (they are extracted from Windows SDK's header files). 

The reason I added them is, before this PR, while we create an instance of `IIterable` in `IVector` and `IVectorView`, we were assuming `IIterable` IID is always the second item on the `iids` list for `IVector` and it is always the last item for `IVectorView` when we call `toInterface()`:
```dart
  // IID_IIterable is always the second item on the iids list for IVector
  late final _iIterable = IIterable<T>.fromRawPointer(toInterface(iids[1]),
      ...);
```
```dart
  // IID_IIterable is always the last item on the iids list for IVectorView
  late final _iIterable = IIterable<T>.fromRawPointer(toInterface(iids.last),
      ...);
```

Unfortunately, I recently learned that this assumption is wrong for some types (oops! :) ). I did some digging and couldn't find a simple way to find the correct `IIterable` IID so I came up with this:

I added these IIDs into the `winrt_constants.dart` and created two `Set`s for them (`mapIterableIids` and `vectorIterableIids`). Then, I created a helper function `iterableIidFromIids` that takes a `List` of interface IIDs and searches between the two `Set`s that I mentioned above to find the correct `IIterable` IID.

With these changes, we now call toInterface() like this:
```dart
  late final _iIterable = IIterable<T>.fromRawPointer(
    toInterface(iterableIidFromIids(iids)),
    ...
  );
```


lib/src/extensions/int_array.dart
```diff
@@ -0,0 +1,84 @@
...
+
+extension Uint8Helper on Pointer<Uint8> {
+  /// Creates a [List] from `Pointer<Uint8>`.
+  ///
+  /// [length] must be equal to the number of elements stored inside the
+  /// `Pointer<Uint8>`.
+  ///
+  /// {@category winrt}
+  List<int> toList({int length = 1}) =>
+      [for (var i = 0; i < length; i++) this[i]];
```
timsneath: Alternative to consider:
```dart
List<int> toList({int length = 1}) => [...this.asTypedList(length)];
```

lib/src/extensions/int_array.dart
```diff
@@ -0,0 +1,84 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// Extension methods to convert integer arrays to List<int>
+
+import 'dart:ffi';
+
+extension Uint8Helper on Pointer<Uint8> {
+  /// Creates a [List] from `Pointer<Uint8>`.
+  ///
+  /// [length] must be equal to the number of elements stored inside the
```
timsneath: Length "must not be greater than", rather than "must be equal to", I think?

lib/src/extensions/int_array.dart
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Extension methods to convert integer arrays to List<int>
+
+import 'dart:ffi';
+
+extension Uint8Helper on Pointer<Uint8> {
+  /// Creates a [List] from `Pointer<Uint8>`.
```
timsneath: Probably worth explaining in the comment why these methods are necessary. That is, of course, these Lists are a copy of the original list that is backed by Dart memory, versus `.asTypedList(length)`, which lives on the C heap and may not remain available.

lib/src/extensions/int_array.dart
timsneath: or another one, perhaps most efficient? 
```dart
List<int> toList({int length = 1}) => Uint8List.fromList(this.asTypedList(length));
```

lib/src/winrt/internal/vector_helper.dart
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  List<int> _toList_int_Uint64() {
+    final pArray = allocator<Uint64>(length);
+    getManyCallback(0, length, pArray);
+    return pArray.toList(length: length);
+  }
+}
+
+const supportedIntTypes = [Int16, Int32, Int64, Uint8, Uint16, Uint32, Uint64];
```
timsneath, line 201: I couldn't help but notice that `Int8` is missing. is that not a viable option? If not, then you might consider a comment with a link to relevant WinRT documentation to avoid confusion.

lib/src/winrt/internal/vector_helper.dart
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  List<T> _toList_COMObject() {
+    final pArray = allocator<COMObject>(length);
+    getManyCallback(0, length, pArray);
+    return pArray.toList(creator!, length: length);
+  }
+
+  List<T> _toList_enum() {
+    switch (intType) {
```
timsneath, line 72: Maybe a comment here to highlight that for enums, the only valid WinRT types are `Int32` or `UInt32`, with a link to https://docs.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system#enums

lib/src/winrt_helpers.dart
```diff
@@ -179,6 +181,53 @@ Pointer<COMObject> CreateActivationFactory(String className, String iid,
...
+  return func is void Function<X extends T>();
+}
+
+/// Determines whether [S] is a subtype of [T] or [T?].
+///
+/// ```dart
+/// isSubtype<Calendar, IInspectable>(); // true
+/// isSubtype<IUnknown, IInspectable>(); // false
+/// ```
+bool isSubtype<S, T>() => <S>[] is List<T> || <S>[] is List<T?>;
```
timsneath, line 201: Do we have tests for these new helpers?

tool/generator/lib/src/inputs/interfaces.dart
```diff
@@ -115,6 +115,7 @@ final windowsRuntimeTypesToGenerate = <String>{
   'Windows.Globalization.Calendar',
   'Windows.Globalization.PhoneNumberFormatting.PhoneNumberFormatter',
   'Windows.Globalization.PhoneNumberFormatting.PhoneNumberInfo',
+  'Windows.Graphics.Printing3D.Printing3DMultiplePropertyMaterial',
```
timsneath, line 118: Did you pick this one just as a test case? (It's fine -- just curious...)

timsneath (commented):
This is truly impressive work. I remain amazed by your technical prowess! I've added a few comments...

tool/generator/lib/src/inputs/interfaces.dart
halildurmus: Yes! This seemed like a great choice as it is `Activatable` so I was able to test it very easily.

lib/src/winrt_helpers.dart
halildurmus: Added tests for them.

lib/src/winrt/internal/vector_helper.dart
halildurmus: Done

lib/src/winrt/internal/vector_helper.dart
halildurmus: `Int8` is not supported by the WinRT Type system: https://docs.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system#fundamental-types

I added a comment with the link above explaining this.

lib/src/extensions/int_array.dart
halildurmus: Done

lib/src/extensions/int_array.dart
halildurmus: Done

EDIT: Since the Lists returned from these methods are `fixed-length` Lists, I made some changes for consistency so that IVector<T>.toList() now returns an immutable List (just like IVectorView's toList()).

lib/src/extensions/int_array.dart
halildurmus: These methods and also these https://github.com/timsneath/win32/blob/b68ef9fbedbada9dcaa7b6994c856001acfa9457/lib/src/extensions/comobject_pointer.dart#L29 https://github.com/timsneath/win32/blob/b68ef9fbedbada9dcaa7b6994c856001acfa9457/lib/src/extensions/hstring_array.dart#L25 are used in VectorHelper's `toList` methods to create a Dart `List` from the pointer filled by the `getMany` methods. 

It is highly unlikely the consumers will need these helpers as they should use `toList()` methods provided by `IVector` and `IVectorView`.

Because of that, I've been thinking maybe we shouldn't expose these extension methods to consumers and move these into the `winrt/internal` folder in order to prevent any confusion that might arise.

timsneath: Yeah, that makes sense. The fewer extension methods exposed on common intrinsic types like `int` or `String`, the better, I think. We should still explain the difference between this and `.asTypedList` though, for someone reading the code who might not immediately recognize why we're copying the list.

lib/src/winrt/internal/vector_helper.dart
timsneath: Great - thanks.

timsneath (approved):
As discussed in the comment, let's not expose the extension types on intrinsics to package consumers if we don't need to, but otherwise LGTM! Thank you!!!

lib/src/extensions/int_array.dart
halildurmus: Great! I pushed some new commits. Can you take another look?

timsneath:
Looks good. Ready to check in. Perhaps as you merge (or as part of a separate pull request), we should move the hidden IIDs in winrt_constants into the internal\ folder also...?

PR #524, opened 2022-07-10
Refactor WinRT method and property declarations

halildurmus:
I moved WinRT projection files into the newly created `winrt` folder and created separate classes for method and property declarations. What do you think?


tool/generator/lib/src/projection/winrt/declarations/comobject.dart
```diff
@@ -0,0 +1,31 @@
+import '../../../../generator.dart';
+
+class COMObjectMethod extends WinRTMethodProjection {
```
timsneath: I'm a bit worried about the name confusion with `ComMethodProjection`. We may need something more precise here, like `WinRTMethodReturningComObjectProjection`. I know it's a long name, but we need some kind of convention.

tool/generator/lib/src/projection/winrt/declarations/comobject.dart
```diff
@@ -0,0 +1,31 @@
...
+        final retValuePtr = calloc<COMObject>();
+        $parametersPreamble
+        ${ffiCall()}
+        $parametersPostamble
+        return retValuePtr;
+      }
+''';
+}
+
+class COMObjectGetProperty extends WinRTGetPropertyProjection {
```
timsneath: Similar to above, confusion between this and `ComGetPropertyProjection`. This would perhaps be `WinRTGetPropertyReturningComObjectProjection`?

tool/generator/lib/src/projection/winrt/declarations/datetime.dart
```diff
@@ -0,0 +1,63 @@
+import '../../../../generator.dart';
+
+class DateTimeMethod extends WinRTMethodProjection {
```
timsneath: Whatever name convention we adopt above for COMObject, we should replicate for other methods.

tool/generator/lib/src/projection/winrt/declarations/comobject.dart
halildurmus: Yeah, you're right. It does cause some confusion with other classes.

I renamed them based on your suggestions (e.g. `WinRTMethodReturningXXProjection`, `WinRTGetPropertyReturningXXProjection`, and `WinRTSetPropertyReturningXXProjection`).

tool/generator/lib/src/projection/winrt/declarations/comobject.dart
halildurmus: Done

tool/generator/lib/src/projection/winrt/declarations/datetime.dart
halildurmus: Done

timsneath (approved):
LGTM!

Issue #461, opened 2022-05-31
Decide what WinRT classes to feature/include for next release


timsneath:
We haven't shipped an updated version to pub.dev for a little while, because there's been lots of (good) churn on the WinRT front. 

Before we ship again, we should decide which WinRT classes we publish/export. Right now, I've been adding classes purely based on whether they help us improve the projection, rather than whether they're _useful_ to package consumers. We're still limited to classes that don't require async. 

[ ] ApplicationData isn't necessary - that was requested when we were exploring a UWP runner for Flutter. 
[X] Calendar and PhoneNumber are quite good examples, even if they're of somewhat useful utility
[X] Gamepad is mostly serviceable, although I need to test it with a real connected controller

What else would be useful to add for this initial wave?

halildurmus:
I'm almost done adding support for asynchronous APIs (need to update the generator to project these APIs as `Future` and write some tests). It should be ready for review in a couple of days.). In fact, I plan to send a draft PR tonight, you might want to take a look. 

Since `Dart` doesn't support async callbacks yet, I'm using `Completer` and polling `IAsyncInfo`'s `Status` until the function completes (with sleeps in the meantime of course). After the async function completes, I call the `GetResults` method and use the result to complete the `Future` with `Completer`'s `complete()` method.  It seems to be working pretty well.

While working on the generator today, I realized that the work I'm going to do would also allow us to project `Pointer<COMObject>` types to their respective types. For example, `ICalendar`'s `Clone()` would return `Calendar` instead of `Pointer<COMObject>`.

> What else would be useful to add for this initial wave?

I don't have anything in my mind right now but after we land these changes, I think we will have a lot of options to choose from 😄.

timsneath:
Things to do to ship a 3.0.0-pre (draft for discussion):

- [x] Move WinRT exports to a separate library/libraries within this package (`win32/winrt.dart`?) just like winsock2 is exposed as a separate library
- [x] Establish the nature of the breaking changes, particularly for COM, and write up a migration guide
- [x] Validate and edit the list of WinRT shipped classes, per comment above
- [x] Improve documentation and examples for WinRT
- [x] Tidy up the README by moving the examples table elsewhere and doing a better job of establishing the value proposition of this package
- [x] Make sure that analyzer lints are resolved in the configuration pana uses them

PR #530, opened 2022-07-15
Support `IMap` and `IMapView`

### LINKED ISSUE #486: Support `IMap` & `IMapView`

halildurmus:
Closes #486

- Adds support for `IKeyValuePair`, `IMap`, and `IMapView` interfaces
- Adds `PedometerStepKind` enum (used as a key type for the above interfaces)
- Adds `PropertySet`, `MediaPropertySet`, `StringMap` and `ValueSet` classes (implementations of the `IMap`)
- Adds `NotificationData` class

After we land this, I'll send a PR that will include support for [Windows.Data.Json](https://docs.microsoft.com/en-us/uwp/api/windows.data.json?view=winrt-22621) and tests for `IMap<String, IJsonValue?>`.

As for the `IMap<Object, Object?>` type, I haven't been able to test it as the [ResourceDictionary](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.resourcedictionary?view=winrt-22621) is the only API that supports it, and I think it only works in `UWP` apps.


lib/src/guid.dart
```diff
@@ -85,6 +85,22 @@ class GUID extends Struct {
...
+    return other is GUID &&
+        other.Data1 == Data1 &&
+        other.Data2 == Data2 &&
+        other.Data3 == Data3 &&
+        other.Data4 == Data4;
+  }
+
+  @override
+  int get hashCode =>
+      Data1.hashCode ^ Data2.hashCode ^ Data3.hashCode ^ Data4.hashCode;
```
halildurmus, line 102: These overrides are required for `Map<GUID, XX>` types (e.g. `Map<GUID, Object?>`) to work properly.

lib/src/winrt/foundation/collections/ikeyvaluepair.dart
```diff
@@ -0,0 +1,278 @@
...
+
+    if (isSubtypeOfWinRTEnum<V>() && enumCreator == null) {
+      throw ArgumentError.notNull('enumCreator');
+    }
+  }
+
+  /// Gets the key of the key-value pair.
+  K get key {
+    if (isSameType<K, GUID>()) return _key_GUID as K;
+    if (isSameType<K, int>()) return _key_Uint32 as K;
```
halildurmus, line 80: I used `Uint32` for `int` keys since it is the only `int` type to be supported for `IKeyValuePair`: https://github.com/timsneath/win32/blob/339c6e4ac04a9d2f2c1933c958bc0893be65dee2/lib/src/winrt_constants.dart#L147-L148

lib/src/winrt/foundation/collections/ikeyvaluepair.dart
```diff
@@ -0,0 +1,278 @@
...
+            .asFunction<int Function(Pointer, Pointer<COMObject>)>()(
+        ptr.ref.lpVtbl, retValuePtr);
+
+    if (FAILED(hr)) throw WindowsException(hr);
+
+    return _creator!(retValuePtr);
+  }
+
+  V get _value_enum {
+    final retValuePtr = calloc<Int32>();
```
halildurmus, line 196: Used `Int32` here for enum values since all supported enum types for `IKeyValuePair` are `Int32` type (e.g. `AppCapabilityAccessStatus`, `ChatMessageStatus`).

lib/src/winrt/foundation/collections/ikeyvaluepair.dart
```diff
@@ -0,0 +1,278 @@
...
+
+      if (FAILED(hr)) throw WindowsException(hr);
+
+      return retValuePtr.value;
+    } finally {
+      free(retValuePtr);
+    }
+  }
+
+  PedometerStepKind get _key_PedometerStepKind {
```
timsneath: Forgive me, but this seems weird here. Why would such a specific type as `PedometerStepKind` be embedded in a generic collection object like this? There could be hundreds of these, couldn't there? What if we decide (as is likely) to partition `Windows.Device.Sensors` into a different export from `Windows.Foundation.Collections`?

lib/src/winrt/foundation/ipropertyvalue.dart
```diff
@@ -38,6 +33,93 @@ class IPropertyValue extends IInspectable {
   factory IPropertyValue.from(IInspectable interface) =>
       IPropertyValue.fromRawPointer(interface.toInterface(IID_IPropertyValue));
 
+  Object? get value {
```
timsneath: I'm wondering if this should be implemented by keeping the interface autogenerated but adding an extension method for this helper method? The more that is autogenerated, the better, obviously -- both less prone to errors and more likely to catch errors in the generation. WDYT?

lib/src/winrt/foundation/ipropertyvalue.dart
```diff
@@ -363,24 +445,18 @@ class IPropertyValue extends IInspectable {
   GUID getGuid() {
```
timsneath: Is this deallocation a bug in the generator code? Or is this fix unique to this interface?

timsneath (commented):
This is an astonishing piece of work. You have outclassed me! I've been busy in my free time fixing up another package that I own which has bitrotted a little, so I'm thrilled to see you continuing to make progress here. A few comments inline to check out. 

At some point, we need to converge on a release, since there are folk waiting on more mundane fixes (e.g. some of the Win32 APIs that have been added over the last couple of months. I'm thinking of shipping a `3.0.0-pre` based on what we've got so far to help out those who are waiting, and then we can keep iterating until you feel the new WinRT projection has settled down. WDYT?

lib/src/winrt/foundation/ipropertyvalue.dart
halildurmus: > Is this deallocation a bug in the generator code? 

Yeah, it is a bug in the generator. `retValuePtr` should not be freed for structs (just like for `COMObject`s).

I don't know how can we avoid memory leaks from these as `NativeFinalizer` cannot be used with structs.

lib/src/winrt/foundation/ipropertyvalue.dart
halildurmus: Yeah, that makes sense. Before doing that, I need to modify the generator to fix the deallocation bug I mentioned below.

lib/src/winrt/foundation/collections/ikeyvaluepair.dart
halildurmus: Here are the supported key-value pairs of `IKeyValuePair` according to IIDs taken from [here](https://github.com/Elmue/WindowsRT-GUID-Analyzer/blob/main/Release/Windows%2011%20(Build%2022000)%20WinRT%20Interfaces.ini):
https://github.com/timsneath/win32/blob/339c6e4ac04a9d2f2c1933c958bc0893be65dee2/lib/src/winrt_constants.dart#L62-L148

According to these, there is only one enum type (`PedometerStepKind`) used as a key type of `IKeyValuePair`. It seemed easier to handle it this way instead of adding a new parameter (`keyCreator`) that takes an enum constructor for the key type of `ÌKeyValuePair` which will allow me to create an instance of that enum type. 

I can modify the implementation to handle these types with the `keyCreator` parameter I mentioned above if that's what you want.

halildurmus:
> This is an astonishing piece of work. You have outclassed me! I've been busy in my free time fixing up another package that I own which has bitrotted a little, so I'm thrilled to see you continuing to make progress here. A few comments inline to check out. 

Thank you!

> At some point, we need to converge on a release, since there are folk waiting on more mundane fixes (e.g. some of the Win32 APIs that have been added over the last couple of months. I'm thinking of shipping a `3.0.0-pre` based on what we've got so far to help out those who are waiting, and then we can keep iterating until you feel the new WinRT projection has settled down. WDYT?

That's a great idea. I saw your [comment](https://github.com/timsneath/win32/issues/461#issuecomment-1187702678) and I think these are great. I'll let you know if I come up with something to add there.

lib/src/winrt/foundation/collections/ikeyvaluepair.dart
timsneath: I wonder if the best thing is to make the `PedometerStepKind` pieces an extension method in winrt/device/sensors? I don't _know_ that this is the right answer, but it does feel like a very odd outlier. Are there other classes that might have a similar pattern, I wonder?

halildurmus: > I wonder if the best thing is to make the `PedometerStepKind` pieces an extension method in winrt/device/sensors?

I've created extension methods for the `PedometerStepKind` pieces but I'm not sure if it's the way you wanted it. Could you take a look?

> Are there other classes that might have a similar pattern, I wonder?

I don't think there are.

lib/src/winrt/foundation/ipropertyvalue.dart
halildurmus: Done

timsneath (approved):
LGTM!

PR #541, opened 2022-08-11
Support creating `null` HSTRING

halildurmus:
[WindowsCreateString function (winstring.h)](https://docs.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowscreatestring#parameters)
> ... To create a new, empty, or NULL string, pass NULL for sourceString and 0 for length.

EDIT: The failed check for formatting in [build(windows-latest, main)](https://github.com/timsneath/win32/runs/7784240786?check_suite_focus=true) might be related to changes made to the `dart format` command in the latest Dart SDK version.

### DIFF
--- a/lib/src/winrt/foundation/collections/imap.dart
+++ b/lib/src/winrt/foundation/collections/imap.dart
@@ -661,7 +661,7 @@ class IMap<K, V> extends IInspectable
   bool _insert_String_String(String key, String? value) {
     final retValuePtr = calloc<Bool>();
     final hKey = convertToHString(key);
-    final hValue = value != null ? convertToHString(value) : 0;
+    final hValue = convertToHString(value);
 
     try {
       final hr = ptr.ref.lpVtbl.value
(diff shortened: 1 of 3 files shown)


timsneath:
Sorry for the delay in reviewing. 

I'm wondering if this is the right way to go after reading the [HSTRING introduction](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring), specifically (emphasis mine):

> The HSTRING handle is a standard handle type. Semantically, an HSTRING containing the value NULL represents the empty string, which consists of zero content characters and a terminating NULL character. Creating a string via [WindowsCreateString](https://docs.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowscreatestring) with zero characters will produce the handle value NULL. When calling [WindowsGetStringRawBuffer](https://docs.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowsgetstringrawbuffer) with the value NULL, a pointer to an empty string followed only by the NUL terminating character will be returned. **No distinct value exists to represent an HSTRING that is uninitialized.**

If I take that at face value, it would suggest that `convertToHString` should require `String`, rather than `String?`, since there is no "null" string, just empty strings. 

What do you think? You're closer than this to me.

halildurmus:
> ...
> If I take that at face value, it would suggest that `convertToHString` should require `String`, rather than `String?`, since there is no "null" string, just empty strings.

Right! I guess I didn't pay enough attention while reading the docs :). Sorry.

PR #542, opened 2022-08-11
Support nullable types (`IReference<T>`)

### LINKED ISSUE #540: Support projection of nullable types (`IReference<T>`)
https://docs.microsoft.com/en-us/uwp/api/windows.foundation.ireference-1?view=winrt-22621

halildurmus:
Closes #540 

> IReference<T> inherits [IPropertyValue](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.ipropertyvalue?view=winrt-22621). Types that implement IReference<T> also implement the interface members of [IPropertyValue](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.ipropertyvalue?view=winrt-22621).

The above statement is from [here](https://docs.microsoft.com/en-us/uwp/api/windows.foundation.ireference-1?view=winrt-22621#interface-inheritance). Unfortunately, that is not correct for some types.

For example, [DecimalFormatter.ParseInt(String)](https://docs.microsoft.com/en-us/uwp/api/windows.globalization.numberformatting.decimalformatter.parseint?view=winrt-22621#windows-globalization-numberformatting-decimalformatter-parseint(system-string)) returns `IReference<int>` (its native type is Int64) and its `iids` property returns `[{4DDA9E24-E69F-5C6A-A0A6-93427365AF2A}, {00000038-0000-0000-C000-000000000046}]` which does not include `IID_IPropertyValue`. 

That's why `IReference<T>` does not inherit from `IPropertyValue` in my implementation.


timsneath (approved):

tool/generator/lib/src/projection/winrt/declarations/reference.dart
```diff
@@ -0,0 +1,104 @@
...
+    final typeProjection = TypeProjection(typeIdentifier);
+    if (typeProjection.isString) return 'String';
+    if (typeProjection.isWinRTEnum) {
+      return lastComponent(typeProjection.typeIdentifier.name);
+    }
+
+    return typeProjection.methodParamType;
+  }
+
+  String get boxValue {
```
timsneath, line 22: Just a note: as the generation is getting more and more sophisticated, with many layers of hierarchy, I think we need more doc comments for some of these methods and properties. We should strive to explain why they exist and how they're used, not merely what the code does, of course.

halildurmus: I agree. I'll send a PR in a few days.

PR #553, opened 2022-08-31
example - Send RAW data to Windows Printer

lcarrasco:
* Example for sending RAW data (eg ESC/POS commands) to a Windows printer using the Winspool API.


example/printer_raw.dart
```diff
@@ -0,0 +1,160 @@
...
+  // Example: ESC/POS sequence to open the cash drawer
+  const openCashDrawer = '\x1b\x70\x00';
+
+  // NOTE: You can get the printer name from the printer_list.dart example
+  final printer = RawPrinter('EPSON TM-T20II Receipt');
+  final data = <String>[];
+
+  for (var i = 0; i < 10; i++) {
+    data.add('Hello world line $i\n');
+  }
```
timsneath: Using a list comprehension here is a little more idiomatic for Dart.

example/printer_raw.dart
```diff
@@ -0,0 +1,160 @@
...
+
+    return res;
+  }
+}
+
+void main() {
+  // Example: ESC/POS sequence to open the cash drawer
+  const openCashDrawer = '\x1b\x70\x00';
+
+  // NOTE: You can get the printer name from the printer_list.dart example
```
timsneath: By using Arena to manage allocations, you can have Dart automatically free memory when RawPrinter falls out of scope.

example/printer_raw.dart
```diff
@@ -0,0 +1,160 @@
...
+      if (fSuccess == 0) {
+        final error = GetLastError();
+        throw Exception(
+            'StartDocPrinter error, status: $fSuccess, error: $error');
+      }
+
+      return phPrinter;
+    } finally {
+      free(pPrinterName);
+      free(pDocInfo);
```
timsneath: Switching to `.toNativeUtf16` to use the Arena allocator

example/printer_raw.dart
```diff
@@ -0,0 +1,160 @@
...
+    return EndPagePrinter(phPrinter.value) > 0;
+  }
+
+  bool _endRawPrintJob(Pointer<HANDLE> phPrinter) {
+    return EndDocPrinter(phPrinter.value) > 0 &&
+        ClosePrinter(phPrinter.value) > 0;
+  }
+
+  bool _printRawData(Pointer<HANDLE> phPrinter, String dataToPrint) {
+    final cWritten = calloc<DWORD>();
```
timsneath: Docs say that this returns a non-zero value for success, not that it's a positive integer.

timsneath (commented):
This looks great -- thank you! I tried to tidy up a few things in the example. Could you take a look and check this still works for your POS hardware? If so, I'll approve and merge.

example/printer_raw.dart
lcarrasco: TIL, thanks.

example/printer_raw.dart
lcarrasco: TIL again, thanks.

example/printer_raw.dart
lcarrasco: Nice catch, thanks.

lcarrasco:
> This looks great -- thank you! I tried to tidy up a few things in the example. Could you take a look and check this still works for your POS hardware? If so, I'll approve and merge.

It works perfectly in my POS hardware, thanks. RTM.

timsneath (approved):
Thanks!

PR #556, opened 2022-09-06
Fix projection of nullable enum parameters

halildurmus:
Also made minor tidy up in the declaration classes and added some documentation comments to better explain some things.


timsneath (approved):
Fantastic! This looks great! Thank you!

PR #558, opened 2022-09-12
Fix 'Format a US number' test

halildurmus:
https://github.com/timsneath/win32/blob/f8705863fd288983ce0891d1c019ac64e8942dea/test/winrt_phonenumberformatter_test.dart#L18

The default constructor of `PhoneNumberFormatter` uses the current default region. Since this test is dependent on the current default region being `US` (which is not the case for me), the test fails on my machine as follows:

format

This PR fixes it by using a `US` PhoneNumberFormatter to format the number.


timsneath (approved):
Doh! So much for being i18n-aware! Thanks for fixing this.

PR #564, opened 2022-10-14
Fix memleak on exceptions

timsneath:
If a WinRT call fails with a non-zero `hr`, we throw an exception. But we don't clear up the `retValuePtr` if it has been allocated, in scenarios where we'd otherwise return it to the user. 

This change attempts to `free` all instances of `retValuePtr` that are allocated when an exception occurs. 

(It also fixes a minor test flake that blocked me on my Windows-on-ARM machine.)


halildurmus (approved):
Nice catch! I also made some minor changes. Please take a look and see if you're happy with the changes.

timsneath:
LGTM!

PR #565, opened 2022-10-14
Add pointer and touch APIs

### LINKED ISSUE #562: Support pen info
Hi,

I would like to use this method but it seems it's missing here.
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getpointerpeninfo

timsneath:
Fixes #562


halildurmus (approved):
Thanks!

Issue #563, opened 2022-10-14
Possible use after free bug?


timsneath:
If we're returning this as an `IMapView`, should we be freeing it? Isn't this a "use after free" bug where the memory is still used after it has been freed? @halildurmus 

https://github.com/timsneath/win32/blob/220c807d765518eb7d0216e59c9a13b0bebcfdce/tool/generator/lib/src/projection/winrt/declarations/map.dart#L110

halildurmus:
> If we're returning this as an `IMapView`, should we be freeing it?

We are not returning that as an `IMapView`. We return it as a `Map` by using `toMap()` helper function:
https://github.com/timsneath/win32/blob/220c807d765518eb7d0216e59c9a13b0bebcfdce/tool/generator/lib/src/projection/winrt/declarations/map.dart#L107

In the `toMap` function, we allocate the `pArray` pointer to hold the elements of the `IMapView` ~~and we do not free that pointer so that it's safe to use when we expose them through a `Map`~~: https://github.com/timsneath/win32/blob/220c807d765518eb7d0216e59c9a13b0bebcfdce/lib/src/winrt/internal/map_helpers.dart#L23-L31

**EDIT:** On second thought, we can also free the `pArray` pointer as it's only used when constructing a `Map` (Line 29, while iterating over to the `IKeyValuePair`'s `key` and `value` getters). After that, we don't need `pArray` at all. Therefore, we can safely free it in a `finally` block.

Because of that, I think it's safe to free the `retValuePtr` since we don't return the `IMapView` object.

timsneath:
Great! I think we're good here then. Thanks!

PR #572, opened 2022-10-23
Fix memleak in the `toMap` helper function

halildurmus:
https://github.com/timsneath/win32/blob/220c807d765518eb7d0216e59c9a13b0bebcfdce/lib/src/winrt/internal/map_helpers.dart#L23-L31

... we can also free the `pArray` pointer as it's only used when constructing a `Map` (Line 29, while iterating over to the `IKeyValuePair`'s `key` and `value` getters). After that, we don't need `pArray` at all. Therefore, we can safely free it in a `finally` block.

_Originally posted by @halildurmus in https://github.com/timsneath/win32/issues/563#issuecomment-1279341820_

### DIFF
--- a/lib/src/winrt/foundation/collections/imap.dart
+++ b/lib/src/winrt/foundation/collections/imap.dart
@@ -758,8 +758,7 @@ class IMap<K, V> extends IInspectable
   /// Creates an unmodifiable [Map] from the current [IMap] instance.
   Map<K, V> toMap() => size == 0
       ? Map.unmodifiable(<K, V>{})
-      : MapHelper.toMap<K, V>(first().getMany,
-          length: size, creator: _iterableCreator);
+      : MapHelper.toMap<K, V>(first(), length: size, creator: _iterableCreator);
 
   late final _iIterable = IIterable<IKeyValuePair<K, V>>.fromRawPointer(
       toInterface(iterableIidFromIids(iids)),
--- a/lib/src/winrt/foundation/collections/imapview.dart
+++ b/lib/src/winrt/foundation/collections/imapview.dart
@@ -440,8 +440,7 @@ class IMapView<K, V> extends IInspectable
   /// Creates an unmodifiable [Map] from the current [IMapView] instance.
   Map<K, V> toMap() => size == 0
       ? Map.unmodifiable(<K, V>{})
-      : MapHelper.toMap<K, V>(first().getMany,
-          length: size, creator: _iterableCreator);
+      : MapHelper.toMap<K, V>(first(), length: size, creator: _iterableCreator);
 
   late final _iIterable = IIterable<IKeyValuePair<K, V>>.fromRawPointer(
       toInterface(iterableIidFromIids(iids)),
--- a/lib/src/winrt/internal/map_helpers.dart
+++ b/lib/src/winrt/internal/map_helpers.dart
@@ -35,6 +35,7 @@ class MapHelper {
       return Map.unmodifiable(map);
     } finally {
       free(pKeyValuePairArray);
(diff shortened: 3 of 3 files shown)


timsneath:
I couldn't figure out from the code where `pArray` is allocated. Does our code allocate it or just receive memory that WinRT owns?  Is that something that we as the caller are responsible for freeing? https://learn.microsoft.com/en-us/uwp/api/windows.foundation.collections.ivectorview-1.getmany?view=winrt-22621 didn't seem to give me a strong clue that we should be freeing it...

halildurmus:
I found another memory leak in the `toMap` methods of `IMap` and `IMapView`: https://github.com/timsneath/win32/blob/f97c950a8d891b4df97dacb548172e552ba55431/lib/src/winrt/foundation/collections/imapview.dart#L440-L444

In Line `443`, we pass `first().getMany` function to `toMap` helper function but we do not free the `IIterator` object returned from the `first` method.

To fix this, I modified the `toMap` function to take the `IIterator` object directly and free its pointer after it's no longer used.

> I couldn't figure out from the code where `pArray` is allocated.

It's allocated here: https://github.com/timsneath/win32/blob/a6fd24d1b20fb9a252723f3483f6b59df956a152/lib/src/winrt/internal/map_helpers.dart#L25 (I renamed it to `pKeyValuePairArray` in this PR)

> Does our code allocate it or just receive memory that WinRT owns?

Yeah, our code allocates it and passes it to the `getMany` function with its length: https://github.com/timsneath/win32/blob/a6fd24d1b20fb9a252723f3483f6b59df956a152/lib/src/winrt/internal/map_helpers.dart#L28

> Is that something that we as the caller are responsible for freeing? https://learn.microsoft.com/en-us/uwp/api/windows.foundation.collections.ivectorview-1.getmany?view=winrt-22621 didn't seem to give me a strong clue that we should be freeing it...

Yes, it's our responsibility to free it. There is a great post from Raymond Chen: [The various patterns for passing C-style arrays across the Windows Runtime ABI boundary](https://devblogs.microsoft.com/oldnewthing/20200205-00/?p=103398)

In the above post, there is a table that summarizes the differences between different pass array styles and also explains who is responsible for freeing the pointer. 

Since the `getMany` function is using the `FillArray` style, it's our responsibility as the caller to free it.

How do I know that the `getMany` function uses the `FillArray` style? Here is the `IIterator`'s `GetMany` implementation from `windows.foundation.collections.h` file:
```h
  template <class U, class T>
  HRESULT _IteratorGetMany(_In_ U* pThis, _In_ unsigned capacity, _Out_writes_to_(capacity,*actual) T *value, _Out_ unsigned *actual)
  {
    HRESULT hr = S_OK;
    ::boolean fHasCurrent = false;
    unsigned count = 0;
    ZeroMemory(value, sizeof(*value) * capacity);
    *actual = 0;

    hr = pThis->get_HasCurrent(&fHasCurrent);
    while (SUCCEEDED(hr) && (fHasCurrent) && (count < capacity))
    {
      hr = pThis->get_Current(&value[count]);
      if (SUCCEEDED(hr))
      {
        count++;
        hr = pThis->MoveNext(&fHasCurrent);
      }
    }

    if (SUCCEEDED(hr))
    {
      *actual = count;
    }
    else
    {
      // cleanup output paremeters on failure
      // no need to zero out *actual as it is still
      // initialized to zero.
      Detail::_Cleanup(value, *actual);
    }
    return hr;
  }
```

You can see from the parameter declaration that it uses the `FillArray` style (see the ABI row in the table from Raymond's post).

timsneath (approved):
LGTM!

PR #575, opened 2022-10-24
Fix memleak on exceptions (in WinRT methods)

halildurmus:
It looks like we forgot to update the `ffiCall` method in the `WinRTMethodProjection` to handle `freeRetValOnFailure` flag in #564 (Oops! 😅).


timsneath (approved):
LGTM!

PR #559, opened 2022-09-13
Fix projection of WinRT object types

halildurmus:
This PR also;
   - Allows more classes and interfaces to be automatically generated
   - Adds `BatteryReport`, `Headset`, `HostName`, `IPInformation`, `NetworkItem` and `NetworkAdapter` classes (some of them required by `Gamepad` and `HostName` classes)
   - Adds `BatteryStatus` and `NetworkTypes` enumerations (required by `Gamepad` and `HostName` classes)
   - Adds support for projection of COMObject setters

PR #576, opened 2022-10-24
Generate code for missing classes

timsneath:
Not sure how, but these classes didn't get generated before.


halildurmus (approved):
The changes came from #575 as expected (#559 was merged before #575, therefore #559 did not have these changes).

timsneath:
Makes sense. Thanks!

PR #578, opened 2022-10-25
Use the relative path from the current folder while generating imports

halildurmus:
Currently, we are using relative paths to the `src` directory when generating imports. With this PR, the relative path from the current folder is used instead.

Take `IGamepad2` for example. It's currently located in `winrt/gaming/input`. It needs to import the `IGamepadController`. As they both are in the same folder, the import is generated as `import 'igamepadcontroller.dart';`.

Another example is `User` (located in `winrt/system`). The relative path from the current folder (`winrt/gaming/input`) should be calculated to the folder where the `User` is located e.g. `import '../../system/user.dart';`

igamepad2

### DIFF
--- a/tool/generator/lib/src/projection/utils.dart
+++ b/tool/generator/lib/src/projection/utils.dart
@@ -1,6 +1,6 @@
 // Useful utilities
 
-import 'package:path/path.dart' as p;
+import 'package:path/path.dart' as path;
 import 'package:winmd/winmd.dart';
 
 import '../shared/exclusions.dart';
@@ -132,9 +132,9 @@ extension CamelCaseConversion on String {
       length >= 2 ? substring(0, 1).toLowerCase() + substring(1) : this;
 }
 
-/// Converts [path] to an equivalent relative path from the [start] directory.
-String relativePath(String path, {required String start}) =>
-    p.relative(path, from: start).replaceAll(r'\', '/');
+/// Converts [targetPath] to an equivalent relative path from the [start] directory.
+String relativePath(String targetPath, {required String start}) =>
+    path.relative(targetPath, from: start).replaceAll(r'\', '/');
 
 String importForWin32Type(TypeIdentifier identifier) {
   if (excludedWin32Structs.contains(identifier.name)) {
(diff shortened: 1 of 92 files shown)


tool/generator/lib/src/projection/utils.dart
```diff
@@ -1,7 +1,6 @@
 // Useful utilities
 
-import 'dart:io';
-
+import 'package:path/path.dart' as p;
```
timsneath: Can we make this `path` rather than `p`? Might be more readable? I think that may be more idiomatic (same way that most people import `math/math.dart` as `math`...

tool/generator/lib/src/projection/winrt/winrt_interface.dart
```diff
@@ -19,8 +17,14 @@ class WinRTInterfaceProjection extends ComInterfaceProjection {
...
 
-  String get relativePathToSrcDir => relativePathToSrcDirectory(File(
-      '../../lib/src/winrt/${folderFromWinRTType(typeDef.name)}/$shortName.dart'));
+  /// Returns the path to the folder where the current interface is located.
+  ///
+  /// e.g. `winrt/storage/pickers`
+  String get _currentFolderPath => 'winrt/${folderFromWinRTType(typeDef.name)}';
+
+  /// Converts [path] to an equivalent relative path from the [_currentFolderPath].
+  String _relativePathTo(String path) =>
```
timsneath: I suspect we don't need to make this function private to the library, since the whole tool is essentially private? Can we just make this `relativePathTo()`? Or is there a reason why this member in particular requires encapsulation?

timsneath (approved):
Superb readability improvement! Thank you. I didn't know about the `.relative()` function in `dart:io` - TIL!

I added a couple of comments to consider, but nothing blocking.

tool/generator/lib/src/projection/utils.dart
halildurmus: Done

tool/generator/lib/src/projection/winrt/winrt_interface.dart
halildurmus: Done

PR #581, opened 2022-10-29
Fix memory leaks in collection interfaces

halildurmus:


lib/src/winrt/foundation/collections/ivector.dart
```diff
@@ -53,21 +52,14 @@ class IVector<T> extends IInspectable implements IIterable<T> {
...
-  /// It is the caller's responsibility to deallocate the returned pointers
-  /// from methods like `GetAt`, `GetView` and `toList` when they are finished
-  /// with it. A FFI `Arena` may be passed as a custom allocator for ease of
-  /// memory management.
   IVector.fromRawPointer(
     super.ptr, {
     T Function(Pointer<COMObject>)? creator,
     T Function(int)? enumCreator,
     Type? intType,
-    Allocator allocator = calloc,
```
halildurmus, line 66: I've removed the `allocator` parameter here (and also in other places) because we don't really use it (aside from a test and manually generated `INetworkInformationStatics` interface). 

Once we start using `NativeFinalizer`, the remaining memory leaks in various places should be freed automatically.

timsneath (approved):
Yay! Cleaner code!

PR #582, opened 2022-10-29
Use metadata for minimum OS version numbers

timsneath:
Merge if/when https://github.com/microsoft/win32metadata/issues/1341 is addressed


timsneath:
This doesn't seem likely to be fixed in the Win32 metadata any time soon, so closing as the branch has significantly diverged from head.

PR #588, opened 2022-11-06
Support `Windows.Foundation.Uri`

### LINKED ISSUE #560: Add `Windows.Foundation.Uri` and expose it to consumers as dart:core's `Uri`
https://docs.microsoft.com/en-us/uwp/api/windows.foundation.uri?view=winrt-22621

halildurmus:
Fixes #560 

It also adds support for `Uri` in collection interfaces (`IIterable`, `IIterator`, `IVector`, and `IVectorView`).


timsneath:
This is great! Do you want to add these to the exports in lib\winrt.dart?

halildurmus:
> This is great! Do you want to add these to the exports in lib\winrt.dart?

I updated the exports but `lib\winrt\foundation\uri.dart` conflicting with dart:core's Uri might be a bit annoying though.

(**Note:** I deleted wwwformurldecodedentry.dart because it has the same contents as the wwwformurldecoderentry.dart. I think I created this file manually while I was working on this and made a typo or something 😄)

timsneath (approved):

timsneath:
> but lib\winrt\foundation\uri.dart conflicting with dart:core's Uri might be a bit annoying though.

Agree with this. Does the WinRT URI class have to be exposed? Or is it fully wrapped by the Dart Uri? I think the approach of other projections is to fully hide WinRT aspects that are wrapped (e.g. some of the `IAsync*` interfaces are wrapped by `async` in C#). We don't have to export everything: just those classes that bleed through outside the interface.

With hindsight, I think my original question was poorly phrased (we should only export those APIs that don't have Dart correspondences). We should probably document our philosophy here.

halildurmus:
> Does the WinRT URI class have to be exposed? Or is it fully wrapped by the Dart Uri?

With this PR, WinRT Uri is fully wrapped by the Dart Uri, so there shouldn't be a need for consumers to access the WinRT Uri.
I think we should not export WinRT Uri and its dependencies (WwwFormUrlDecoder etc.).

> I think the approach of other projections is to fully hide WinRT aspects that are wrapped (e.g. some of the `IAsync*` interfaces are wrapped by `async` in C#). We don't have to export everything: just those classes that bleed through outside the interface.
With hindsight, I think my original question was poorly phrased (we should only export those APIs that don't have Dart correspondences). We should probably document our philosophy here.

I agree.

timsneath:
OK, then my original feedback was in error. Apologies. I was assuming that other exports beyond Uri would be of value, but it seems that we should hide all these for now. 

If you're able to revert this part of the change, I will publish the resultant hash as win32 3.11 (the magic number for any Windows version :) 

Apologies again for the flip-flopping; we're now aligned!

PR #596, opened 2022-11-12
Support generating `IID`s for parameterized types

halildurmus:
The main purpose of this PR is to get rid of the expensive runtime calls in collection interfaces (`iterableIidFromIids` function) and `IReference` interface (`referenceIidFromIids` function).

With this PR, `IIterable` and `IReference` IIDs are generated at compile-time and passed to the relevant interfaces as arguments.
This should greatly improve the performance while using these interfaces.

I recently discovered that these IIDs can be calculated. If I had known this earlier, I wouldn't have created the `iterableIidFromIids` and `referenceIidFromIids` functions in the first place.

Here is a link to documentation explaining how this works https://learn.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system#guid-generation-for-parameterized-types


tool/generator/lib/src/projection/utils.dart
```diff
@@ -341,6 +345,217 @@ String parseGenericTypeIdentifierName(TypeIdentifier typeIdentifier) {
...
+    ..[1] = sha1Bytes[5];
+  final secondNumber = second.buffer.asByteData().getUint16(0);
+
+  final third = Int8List(2)
+    ..[0] = sha1Bytes[6]
+    ..[1] = sha1Bytes[7];
+  var thirdNumber = third.buffer.asByteData().getUint16(0);
+  thirdNumber = (thirdNumber & 0x0fff) | (5 << 12);
+
+  final fourthNumber = (sha1Bytes[8] & 0x3f) | 0x80;
```
halildurmus: These parts are heavily inspired by the Rust/WinRT projection https://github.com/microsoft/windows-rs/blob/1df6f66312ea7cb1e5e08eb7f9f0522b364cc267/crates/libs/windows/src/core/guid.rs#L44C53-L59

tool/generator/lib/src/projection/utils.dart
```diff
@@ -341,6 +345,217 @@ String parseGenericTypeIdentifierName(TypeIdentifier typeIdentifier) {
...
+  return (a > 9) ? a - 10 + 0x61 : a + 0x30;
+}
+
+int _hexsToChars(List<int> guidChars, int offset, int a, int b) {
+  guidChars[offset++] = _hexToChar(a >> 4);
+  guidChars[offset++] = _hexToChar(a);
+  guidChars[offset++] = _hexToChar(b >> 4);
+  guidChars[offset++] = _hexToChar(b);
+  return offset;
+}
```
halildurmus: These are adapted to Dart from the [System.Guid](https://learn.microsoft.com/en-us/dotnet/api/system.guid?view=dotnet-uwp-10.0)'s toString() implementation https://referencesource.microsoft.com/#mscorlib/system/guid.cs,1278

tool/generator/lib/src/projection/utils.dart
```diff
@@ -341,6 +345,217 @@ String parseGenericTypeIdentifierName(TypeIdentifier typeIdentifier) {
...
+        parseTypeIdentifierSignature(typeIdentifier.typeArg!);
+    final secondArgSignature =
+        parseTypeIdentifierSignature(typeIdentifier.typeArg!.typeArg!);
+    return 'pinterface($parentTypeGuid;$firstArgSignature;$secondArgSignature)';
+  }
+
+  return 'pinterface($parentTypeGuid;${parseTypeIdentifierSignature(typeIdentifier.typeArg!)})';
+}
+
+const wrtPinterfaceNamespace = [
```
timsneath, line 470: What do these numbers represent?

halildurmus: They are byte representation of Guid `11f47ad5-7b73-42c0-abae-878b1e16adee`. 

wrt_pinterface_namespace
It is defined in https://learn.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system#guid-generation-for-parameterized-types

tool/generator/lib/src/projection/utils.dart
timsneath: We also have https://github.com/timsneath/win32/blob/196640466804beb70dc2ce1096ef1b17ccc3a7fd/lib/src/guid.dart -- is there scope to merge these two?

halildurmus:
```dart
  ...

  thirdNumber = (thirdNumber & 0x0fff) | (5 << 12);

  final fourth = Int8List(8)
    ..[7] = (sha1Bytes[8] & 0x3f) | 0x80
    ..[6] = sha1Bytes[9]
    ..[5] = sha1Bytes[10]
    ..[4] = sha1Bytes[11]
    ..[3] = sha1Bytes[12]
    ..[2] = sha1Bytes[13]
    ..[1] = sha1Bytes[14]
    ..[0] = sha1Bytes[15];
  final fourthNumber = fourth.buffer.asByteData().getUint64(0);

  final guid = calloc<GUID>();

  try {
    guid.ref.Data1 = firstNumber;
    guid.ref.Data2 = secondNumber;
    guid.ref.Data3 = thirdNumber;
    guid.ref.Data4 = fourthNumber;
    return guid.ref.toString();
  } finally {
    free(guid);
  }
```

This seems to be working. Is that what you mean (use the GUID's toString() to generate the IID)?

timsneath (approved):
I tidied up some of the byte manipulation code a little. PTAL and check you're happy with it. I'd like to rationalize some of the GUID bit twiddling, but let's get this in and then we can always futz with it a bit more!

halildurmus:
Nice cleanup! 🎉

PR #597, opened 2022-11-16
Support generating implements mappers for generic interfaces

### LINKED ISSUE #480: Support generic collections in `implementsMappers`
Noticed this while working on #479. Take `IXmlNodeList` for example. It requires two interfaces: `IVectorView<IXmlNode>` and `IIterable<IXmlNode>`.

The problem is that the `typeDef` of these interfaces returns an empty `name`. This is breaking a lot of things, as we rely heavily on `typeDef.name` for `classDeclaration` and `implementsMappers`.  I ignored them for now in #479.

In order to get the names of these types, we can call typeDef like this:
```dart
print(typeDef.interfaces.first.typeSpec?.type?.name); // Prints Windows.Foundation.Collections.IVectorView`1
print(typeDef.interfaces.first.typeSpec?.typeArg?.name); // Prints Windows.Data.Xml.Dom.IXmlNode
```
The difficulty is that these types can be nested and/or has multiple `typeArg`s (e.g. `IVectorView<IMapView<String, Object>>`, `IMap<String, IVector<String>>`).

In a couple of days, I'll continue my work on #465. In that, I'll have to modify the generator to expose `Pointer<COMObject>` types to their respective types (e.g. IFileOpenPicker's `pickMultipleFilesAsync()` will be exposed as `Future<List<StorageFile>?>` instead of `Pointer<COMObject>` or Calendar's `Clone()` will return `Calendar`). This will probably help to support this as well.

halildurmus:
Fixes #480 

- Refactored map and vector declarations by adding new utility methods for calculating iterableIid.
- Added `IPropertySet` and `IObservableMap` interfaces.
- Added support to generate implements mappers for generic interfaces.
- Extracted logic for mappers projection from `winrt_class.dart` and `winrt_interface.dart` files and created new classes for them (`WinRTFactoryInterfaceMapperProjection`, `WinRTStaticInterfaceMapperProjection`, `WinRTImplementsMapperProjection`)


timsneath (approved):
Looking great! I'm starting to lose track of the inheritance chain, so I'm going to try and map this out somewhere (perhaps a diagram).

halildurmus:
> Looking great! I'm starting to lose track of the inheritance chain, so I'm going to try and map this out somewhere (perhaps a diagram).

Thanks! Yeah, it's starting to get complicated as the generator becomes more sophisticated.

PR #599, opened 2022-11-24
Clean up GUID implementations

timsneath:
Refactors and removes a bunch of fairly hacky code into a cleaner GUID implementation. 

This change introduces a managed, immutable Dart `Guid` type, including various factory constructors and converters to string and `Pointer<GUID>` types. 

It also harmonizes GUIDs to be represented in lower case, which matches the UUID standard.


halildurmus:
This is great work! We should also wrap native GUIDs with Dart `Guid`s (in parameters and return types). I can make the necessary changes.

halildurmus (approved):

PR #602, opened 2022-11-24
Wrap WinRT representations of GUID as Dart Guid

### LINKED ISSUE #601: Wrap WinRT representations of GUID as Dart `Guid`
We should also make sure that we're not using GUID as a key in maps (for example, the `fromMap` test in winrt_collections_test.dart). 

As a test, we should move the `operator==` and `hashCode` methods from the `GUID` class to the `Guid` class. Ideally `GUID` should just be a pure struct, and we should convert it to a Dart `Guid` whenever we want to manipulate or store it.

/cc @halildurmus

halildurmus:
Fixes #601

### DIFF
--- a/lib/src/guid.dart
+++ b/lib/src/guid.dart
@@ -148,8 +148,10 @@ class Guid {
     return true;
   }
 
+  // toString()'s hashCode is used instead of bytes' because the bytes' hashCode
+  // does not work while using Guid as a key in maps.
   @override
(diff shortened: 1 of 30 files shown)


timsneath:
Oh, this is lovely! I'm so glad we made this change -- lots of cleaner code! Thank you 💟

timsneath (approved):

PR #608, opened 2022-11-25
Don't return null for empty strings

halildurmus:
Don't return null for empty strings as there is no `null` string, just empty strings as described in [HSTRING introduction](https://learn.microsoft.com/en-us/windows/win32/winrt/hstring).


timsneath (approved):
LGTM!

PR #611, opened 2022-11-26
Add doc comments for COM and WinRT classes

timsneath:
Moves some more inputs into JSON data format, and adds doc comments for COM and WinRT types.


halildurmus (approved):
Awesome!

PR #613, opened 2022-11-26
Don't generate imports for `XXEventArgs`

halildurmus:
Don't generate imports for `XXEventArgs` classes as they prevent us from autogenerating some interfaces (e.g. `UISettings`).


timsneath (approved):
super!

PR #615, opened 2022-11-27
Make `String` and `WinRTEnum` type arguments non-nullable in WinRT maps

halildurmus:
There is no null string, just empty strings. As for enums, in WinRT, they are represented by the types `Int32` or `UInt32` which defaults to `0`, so there is no null enum either.


timsneath (approved):
Nice! Thank you, as ever!

PR #616, opened 2022-12-01
Sort imports

halildurmus:
With this PR, imports are sorted according to [Effective Dart: Style](https://dart.dev/guides/language/effective-dart/style#ordering) guidelines. I also cleaned up the projection classes a bit.


tool/generator/lib/src/projection/utils.dart
```diff
@@ -557,6 +557,54 @@ String stripGenerics(String name) {
...
 
+/// Sorts [importLines] according to Effective Dart: Style guidelines.
+/// See https://dart.dev/guides/language/effective-dart/style#ordering
+List<String> sortImports(List<String> importLines) {
+  if (importLines.isEmpty) return importLines;
+
+  final dartImports = <String>[];
+  final packageImports = <String>[];
+  final projectImports = <String>[];
+  final projectRelativeImports = <String>[];
```
timsneath: Should these four lines be `Set`s? Presumably we don't want duplicates...

timsneath:
Neat and tidy! I love it :) Thank you.

timsneath (approved):

tool/generator/lib/src/projection/utils.dart
halildurmus: `Set` doesn't have a `sort` method so I've used `SplayTreeSet` which sorts them automatically. Thanks for catching this!

PR #621, opened 2022-12-17
Add SetDisplayConfig

### LINKED ISSUE #618: Add support for setting display config to cloned / extended
I am porting an app from Windows to Flutter (for Windows), and would like to be able to set display config on startup to cloned desktop (same on both screen) or extended desktop.

As far as my own research goes, I think it is [SetDisplayConfig](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setdisplayconfig) that must be used. Afaik it is not implemented yet in the win32 pacakge.

Is it possible to do this with the win32 package?

timsneath:


halildurmus (approved):

timsneath:
Do you ever sleep, @halildurmus?! :) Thank you.

PR #623, opened 2022-12-17
Automatically dispose COM and WinRT objects with `NativeFinalizer`

### LINKED ISSUE #443: Memory management for WinRT Classes
_Originally posted by @halildurmus in https://github.com/timsneath/win32/issues/431#issuecomment-1133899087_

Assume that we're using the `Mixin` model for generating WinRT classes.

There are some cases where we can't deallocate pointers in the generated code. Therefore, it is the consumers' responsibility to deallocate them once they're done using them.

These are:
- WinRT class pointers e.g. `Calendar`'s pointer (these pointers are allocated by constructors such as `Calendar`'s default constructor, factory constructors, methods, properties, or by consumers using `ActivateObject` or other helper methods)
- WinRT mixin pointers e.g. pointers of the `ICalendar` and `ITimeZoneOnCalendar`
- APIs that return `IVector<T extends IInspectable>` or `IVectorView<T extends IInspectable>` e.g. `IVector<StorageFile>`. As of now, these are exposed as `Pointer<COMObject>`. We only auto-generate APIs that return `IVector<String>` and `IVectorView<String>` at the moment but we will add support for these in the future. [Here](https://github.com/timsneath/win32/blob/main/lib/src/winrt/inetworkinformationstatics.dart#L122-L155) is the one I manually generated. 

...and probably much more in the future as the WinRT projection matures (asynchronous APIs, IMap & IMapView...).

The problem is that there are many situations where the consumers need to free pointers manually and I think it is pretty common to forget freeing some of the pointers (especially if you're using a lot of them). Another thing that is quite painful to deal with is that if you accidentally call `free` method twice on the same pointer, the process will crash and good luck figuring out the source of the problem (I got bitten by this many times :( ).

I've been trying to find a way to fix these issues for the past few days and I have some ideas I'd like to share with you.

I'll start with the first design. We could consider introducing a class that manages the pointers that could not be freed in the generated code:
```dart
class WinRTAllocator {
  static final allocator = Arena();

  /// Releases all resources managed by the [allocator].
  static void freeAll() => allocator.releaseAll(reuse: true);
}
```

This class would contain a `static` allocator to be used by the generated code and a static method to free all pointers. We would expose this to the outside world so that the consumers would have a way of freeing all pointers by calling just a single method once they're done using WinRT APIs.

Here are the cases where we need to modify the generator to use this allocator:
- WinRT class pointers
- WinRT mixin pointers (We would pass `WinRTAllocator.allocator` to `toInterface` method as a parameter)
- APIs that return `IVector<T extends IInspectable>` or `IVectorView<T extends IInspectable>`

...and much more in the future...

This design is much simpler to use, but if consumers use WinRT APIs heavily, I'm worried that the memory usage may be higher until they call `WinRTAllocator.freeAll()` (there may be some cases where the consumer is using tightly coupled APIs and because of this they can't call `freeAll` method between calls and must wait for all calls to complete). 

To solve this, here is the final design I came up with. First, we could create our own `Arena` implementation by slightly modifying `dart:ffi`'s implementation to allow freeing specific pointers from the `_managedMemoryPointers` list (by modifying `Arena`'s `free` method):
```dart
class WinRTAllocator {
  static final allocator = _Arena();

  /// Releases memory allocated on the native heap.
  static void free(Pointer<NativeType> pointer) => allocator.free(pointer);

  /// Releases all resources managed by the [allocator].
  static void freeAll() => allocator.releaseAll(reuse: true);
}

class _Arena implements Allocator {
  /// The [Allocator] used for allocation and freeing.
  final Allocator _wrappedAllocator;

  /// Native memory under management by this [Arena].
  final List<Pointer<NativeType>> _managedMemoryPointers = [];

  /// Callbacks for releasing native resources under management by this [Arena].
  final List<void Function()> _managedResourceReleaseCallbacks = [];

  bool _inUse = true;

  /// Creates a arena of allocations.
  ///
  /// The [allocator] is used to do the actual allocation and freeing of
  /// memory. It defaults to using [calloc].
  _Arena([Allocator allocator = calloc]) : _wrappedAllocator = allocator;

  /// Allocates memory and includes it in the arena.
  ///
  /// Uses the allocator provided to the [Arena] constructor to do the
  /// allocation.
  ///
  /// Throws an [ArgumentError] if the number of bytes or alignment cannot be
  /// satisfied.
  @override
  Pointer<T> allocate<T extends NativeType>(int byteCount, {int? alignment}) {
    _ensureInUse();
    final p = _wrappedAllocator.allocate<T>(byteCount, alignment: alignment);
    _managedMemoryPointers.add(p);
    return p;
  }

  /// Registers [resource] in this arena.
  ///
  /// Executes [releaseCallback] on [releaseAll].
  ///
  /// Returns [resource] again, to allow for easily inserting
  /// `arena.using(resource, ...)` where the resource is allocated.
  T using<T>(T resource, void Function(T) releaseCallback) {
    _ensureInUse();
    releaseCallback = Zone.current.bindUnaryCallback(releaseCallback);
    _managedResourceReleaseCallbacks.add(() => releaseCallback(resource));
    return resource;
  }

  /// Registers [releaseResourceCallback] to be executed on [releaseAll].
  void onReleaseAll(void Function() releaseResourceCallback) {
    _managedResourceReleaseCallbacks.add(releaseResourceCallback);
  }

  /// Releases all resources that this [Arena] manages.
  ///
  /// If [reuse] is `true`, the arena can be used again after resources
  /// have been released. If not, the default, then the [allocate]
  /// and [using] methods must not be called after a call to `releaseAll`.
  ///
  /// If any of the callbacks throw, [releaseAll] is interrupted, and should
  /// be started again.
  void releaseAll({bool reuse = false}) {
    if (!reuse) {
      _inUse = false;
    }
    // The code below is deliberately wirtten to allow allocations to happen
    // during `releaseAll(reuse:true)`. The arena will still be guaranteed
    // empty when the `releaseAll` call returns.
    while (_managedResourceReleaseCallbacks.isNotEmpty) {
      _managedResourceReleaseCallbacks.removeLast()();
    }
    for (final p in _managedMemoryPointers) {
      _wrappedAllocator.free(p);
    }
    _managedMemoryPointers.clear();
  }

  /// Releases memory allocated on the native heap.
  @override
  void free(Pointer<NativeType> pointer) {
    final isRemoved = _managedMemoryPointers.remove(pointer);
    if (isRemoved) _wrappedAllocator.free(pointer);
  }

  void _ensureInUse() {
    if (!_inUse) {
      throw StateError(
          'Arena no longer in use, `releaseAll(reuse: false)` was called.');
    }
  }
}
```

(I'd like to point out that calling `WinRTAllocator`'s `free` multiple times on the same pointer won't crash your process as the implementation checks if the pointer exists in the `_managedMemoryPointers` list before calling the native API.)

Then, we could create a `dispose` method in `IInspectable`:
```dart
class IInspectable extends IUnknown {
  ...

  /// Releases all memory allocated by this object.
  void dispose() => WinRTAllocator.free(ptr);
}
```

...and modify WinRT classes and mixins like this:
```dart
// icalendar.dart
mixin ICalendar on IInspectable {
  late final ptrICalendar =
      toInterface(IID_ICalendar, allocator: WinRTAllocator.allocator);
  // ...
}

// itimezoneoncalendar.dart
mixin ITimeZoneOnCalendar on IInspectable {
  late final ptrITimeZoneOnCalendar =
      toInterface(IID_ITimeZoneOnCalendar, allocator: WinRTAllocator.allocator);
  // ...
}

// calendar.dart
class Calendar extends IInspectable with ICalendar, ITimeZoneOnCalendar {
  Calendar() : super(ActivateWinRTClass(_className));
  Calendar.fromPointer(super.ptr);

  static const _className = 'Windows.Globalization.Calendar';

  // ...

  @override
  void dispose() {
    WinRTAllocator.free(ptr);
    WinRTAllocator.free(ptrICalendar);
    WinRTAllocator.free(ptrITimeZoneOnCalendar);
  }
}
```

With this design, consumers have the flexibility to use the `dispose` method of the WinRT classes in combination with the `freeAll` method of the `WinRTAllocator` for memory management. They can decide which option is more suitable for their needs or preferences. Some might prefer to just use `freeAll` for its simplicity while others may use both of them to make their applications more efficient.

halildurmus:
Fixes #443 

Attaches `NativeFinalizer` to `IUnknown` objects so that their pointers can be freed automatically by the NativeFinalizer. Note that freeing their pointers is not enough, we actually need to call `release()` on objects before freeing their pointers so that the resources on the native side can be freed as well. For example, [here](https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/core/unknown.rs#L36-L42) is a snippet from [windows-rs](https://github.com/microsoft/windows-rs) that implements the [Drop](https://doc.rust-lang.org/rust-by-example/trait/drop.html) trait for IUnknown interface.

In order to do this automatically with the NativeFinalizer, we need to write a small function in C++ and use it as the NativeFinalizer's callback. The function is going to be something like this:
```cpp
#include <windows.h>

#define EXPORT extern "C" __declspec(dllexport)

struct COMObject
{
	LPVOID lpVtbl;
};

EXPORT void DisposeComObject(COMObject *com_object)
{
	IUnknown *unknown = reinterpret_cast<IUnknown *>(com_object->lpVtbl);
	unknown->Release();
	CoTaskMemFree(com_object);
}
```

Bundling native code with Dart packages is a huge pain at the moment. Fortunately, there is an upcoming [feature](https://github.com/dart-lang/sdk/issues/50565) to make this process seamless. Once it is released, we should be able to easily use the above function as a callback for the NativeFinalizer and get rid of the calls to `release()`, and make disposing COM and WinRT objects completely automatic (_yay!_).

Also made some refactor/cleanup in various places.


lib/winrt.dart
```diff
@@ -37,28 +37,12 @@
...
-/// integer value, just like other `HANDLE` objects in the Win32 programming
-/// interface.
-///
-/// Helper functions exist to easily convert between the Dart `String` type and
-/// Windows Runtime strings: specifically, [convertToHString] and
-/// [convertFromHString].
-///
-/// Make sure you dispose of `HSTRING`s by calling `WindowsDeleteString`; you do
-/// not need to free the pointer itself, since Windows reference counts the
-/// backing store and frees the memory when the reference count reaches 0.
```
halildurmus, line 60: I removed this section because I think the consumers don't need to know about it since `HSTRING`s are completely wrapped with Dart `String`s.

lib/src/com/iunknown.dart
```diff
@@ -1,25 +1,21 @@
...
-import '../constants.dart';
 import '../exceptions.dart';
 import '../guid.dart';
 import '../macros.dart';
-import '../structs.g.dart';
 import '../utils.dart';
-import '../variant.dart';
-import '../win32/ole32.g.dart';
+
+typedef WinCoTaskMemFreeNative = Void Function(Pointer pv);
```
timsneath: Should this typedef and the final statements below be moved into the class as `static` private members, so they are neatly encapsulated with their usage?

lib/src/com/iunknown.dart
```diff
@@ -33,15 +29,24 @@ const IID_IUnknown = '{00000000-0000-0000-c000-000000000046}';
...
 /// {@category Interface}
 /// {@category com}
-class IUnknown {
+class IUnknown implements Finalizable {
   // vtable begins at 0, is 3 entries long.
   Pointer<COMObject> ptr;
 
-  IUnknown(this.ptr);
+  IUnknown(this.ptr) {
+    _finalizer.attach(this, ptr.cast(), detach: this, externalSize: 8);
```
timsneath: Consider `externalSize: sizeOf<IntPtr>()` to make it more explanatory what this represents?

lib/src/com/iunknown.dart
```diff
@@ -71,16 +81,16 @@ class IUnknown {
   ///
   /// Takes a string (typically a constant such as `IID_IModalWindow`) and does
   /// a COM QueryInterface to return a reference to that interface. This method
-  /// reduces the boilerplate associated with calling QueryInterface manually.
+  /// reduces the boilerplate associated with calling `queryInterface` manually.
   Pointer<COMObject> toInterface(String iid) {
-    final pIID = convertToIID(iid);
-    final pObject = calloc<COMObject>();
+    final nativeGuidPtr = convertToIID(iid);
```
timsneath: Curious about this change. From a documentary perspective, this is an IID, shouldn't we name it as such?

lib/winrt.dart
timsneath: Great!

test/winrt_calendar_test.dart
```diff
@@ -329,15 +334,6 @@ void main() {
       expect(resolvedLanguage.length, equals(5));
     });
 
-    test('Calendar.numeralSystem getter', () {
```
timsneath, line 332: Curious -- is there a reason why this test is not good?

timsneath (commented):
A few comments prior to merging; hopefully they're helpful!

test/winrt_calendar_test.dart
halildurmus: This was a duplicate of the test in Line 63.

lib/src/com/iunknown.dart
halildurmus: Yeah, you're right. I reverted the change.

lib/src/com/iunknown.dart
halildurmus: Done

lib/src/com/iunknown.dart
halildurmus: That makes more sense, thanks!

timsneath (approved):
Great! Let's do this! :) Thank you.

PR #628, opened 2022-12-24
Explore typedefs

timsneath:
Exploring whether the use of the Dart typedefs materially improves the readability of the generated code.


timsneath:
I'm going to abandon this pull request, unfortunately. It doesn't improve the experience of a package consumer, since only the Dart types are exposed through IntelliSense. And the win32metadata project doesn't make this easy to sustain, since they're replacing some traditional Win32 types with annotated types that don't translate well in this format.

Keeping this around in case there's a future reason to resurrect this code, but I'm going to close it for now as unplanned.

PR #635, opened 2023-01-01
Fix projection of async methods that return IAsyncOperation delegate

halildurmus:
Before this change:
```dart
    final asyncOperation =
        IAsyncOperation<GeolocationAccessStatus>.fromRawPointer(retValuePtr,
            creator: GeolocationAccessStatus.from);
```

After this change:
```dart
    final asyncOperation =
        IAsyncOperation<GeolocationAccessStatus>.fromRawPointer(retValuePtr,
            enumCreator: GeolocationAccessStatus.from, intType: Int32);
```


timsneath (approved):
Great!

PR #638, opened 2023-01-02
Generalize map collection interfaces to handle all enum type arguments

halildurmus:
Fixes:

> The changes to the core package are a little more painful: mostly because it introduces a circular dependency, since IMapView and other foundation classes use a couple of classes from Windows.Devices (`BatteryReport` and `PedometerStepKind`, most prominently). Ideally we wouldn't have to depend on the child packages from the foundation package, but I don't know if that's possible.

_Originally posted by @timsneath in https://github.com/timsneath/win32/discussions/634#discussioncomment-4569974_


timsneath (approved):
Oh, that's great! Thank you!

PR #642, opened 2023-01-04
Support automatic initialization of the WinRT and remove `winrtUnitialize` helper function

### LINKED ISSUE #643: Segfault with WinRT hosting
> This was a segfault, where Dart just quits without warning, rather than a process exit as in #463. The last displayed test attempt was `IReference<Int64>`. Until the changes I made here, Dart tests ran with the default concurrency settings (two threads), so there's a 1 in 2 chance that it was this specific test that failed. But both tests running would have been from that test file, I think.

I came across this today. You should be able to reproduce this with the following cmd script:
```cmd
@echo off

rem Run the dart test command 30 times
for /l %%i in (1,1,30) DO (
  echo Run: %%i
  call dart test test/winrt_calendar_test.dart test/winrt_collections_test.dart test/winrt_ireference_test.dart
)
```
(You may need to run this script several times 😄)

While running this script, you should see either a segfault or an error like this in some random test:
```
  Error 0x80080008: Object server is stopping when OLE service contacts it
  package:win32/src/winrt_helpers.dart 140:21                ActivateClass
  package:win32/src/winrt/globalization/calendar.dart 37:15  new Calendar
  test\winrt_calendar_test.dart 18:18                        main.<fn>
```

I think this is caused by calls to the `winrtUninitialize()` helper function which calls `RoUninitialize()`. I can't reproduce this if I comment out the call to `RoUninitialize()` in the winrtUninitialize().

I have looked at the other WinRT projections to see if they call `RoUninitialize()` and I can confirm that they don't. In fact, they don't call `RoInitialize()` either. They support automatic initialization of the Windows Runtime by using [CoIncrementMTAUsage](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coincrementmtausage) API.

Here is the relevant snippet from `cppwinrt`:
https://github.com/microsoft/cppwinrt/blob/fe304096fa30583f3cc2ebfdbf564d2b4081e2ad/strings/base_activation.h#L37-L52

And this is from `windows-rs`:
https://github.com/microsoft/windows-rs/blob/a8b3b3841ae5f966dd5f86397a0a81eaa26470b6/crates/libs/windows/src/core/factory_cache.rs#L56-L73

Basically, if the call to `RoGetActivationFactory` fails with the error code `CO_E_NOTINITIALIZED`, they initialize a Multi-Threaded Apartment (MTA) on the current thread by calling `CoIncrementMTAUsage()` first and then try again.

I think we should do the same, Created #642 for this. Curious to hear your thoughts.

Note that I forgot to mention in #623, before we make NativeFinalizer call IUnknown's `release()` automatically on objects we would need to stop calling `winrtUninitialize()` because, by the time NativeFinalizer callbacks run, the WinRT would be uninitialized which would crash the app (I have experienced this multiple times while working on #623). So, even if we didn't have this problem, we would have to stop using `winrtUninitialize` at some point.

_Originally posted by @halildurmus in https://github.com/timsneath/win32/issues/630#issuecomment-1371308755_

halildurmus:
Context: #643

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -8,6 +8,8 @@ require Flutter.
 | `bluetooth.dart`      | Demonstrate enumerating Bluetooth devices                 |
 | `bluetoothle.dart`    | Enumerate Bluetooth LE (Low Energy) devices               |
 | `calendar.dart`       | Gets information about the calendar from a WinRT API      |
+| `com_context.dart`    | Shows interaction of Dart isolates and COM apartments     |
+| `com_demo.dart`       | Demonstrates COM object reference counting                |
 | `commdlg.dart`        | Demonstrates using the color chooser common dialog box    |
 | `credentials.dart`    | Adds a credential to the store and retrieves it           |
 | `customtitlebar.dart` | Demonstrates creation of owner-draw title bar region      |
--- a/example/appcontainer.dart
+++ b/example/appcontainer.dart
@@ -40,8 +40,6 @@ bool isAppContainer() {
 }
 
 void main() {
-  winrtInitialize();
-
   print('${!isAppContainer() ? '!' : ''}isAppContainer');
 
   final userData = UserDataPaths.getDefault();
--- a/example/appcontainer.dart
+++ b/example/appcontainer.dart
@@ -45,6 +45,4 @@ void main() {
   final userData = UserDataPaths.getDefault();
   final roamingAppData = userData.roamingAppData;
   print('RoamingAppData: $roamingAppData');
-
-  winrtUninitialize();
 }
--- a/example/calendar.dart
+++ b/example/calendar.dart
@@ -15,7 +15,6 @@ String calendarData(Calendar calendar) =>
     'Time Zone: ${calendar.timeZoneAsFullString()}\n';
 
 void main() {
-  winrtInitialize();
   try {
     print('Windows Runtime demo. Calling Windows.Globalization.Calendar...\n');
     final calendar = Calendar();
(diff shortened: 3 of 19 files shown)


timsneath (approved):
Thanks! I've added an example to help myself comprehend how Dart isolates interact with the COM apartment threading model, using @kennykerr's [blog as a guide](https://web.archive.org/web/20221128052910/https://kennykerr.ca/2018/04/03/cppwinrt-understanding-coroutines-and-the-calling-context/). 

I also tidied up the `_initializeMTA` function and added a couple of Win32 functions to support the example.

Please take a look at my additional changes and make sure you're happy with my work before you merge it!

halildurmus:
Love it, thank you so much!

timsneath:
One last thought, @halildurmus -- I wonder if we shouldn't leave both `winrtInitialize()` and `winrtUninitialize()` as empty (but deprecated) functions, just for compatibility reasons? Otherwise we have to update the major version number again. 

I noticed that you removed the latter but kept the former - was that intentional?

halildurmus:
> One last thought, @halildurmus -- I wonder if we shouldn't leave both `winrtInitialize()` and `winrtUninitialize()` as empty (but deprecated) functions, just for compatibility reasons? Otherwise we have to update the major version number again.

 Ah, you're right. I'm away from my computer. Feel free to restore the `winrtUninitialize()` function and deprecate it, or I'll take care of it later today.

> I noticed that you removed the latter but kept the former - was that intentional?

Yes. There are a few APIs that only work in Single-Threaded Apartment (STA). `winrtInitialize()` should be used in these cases (although I couldn't find any API to test this).

PR #645, opened 2023-01-05
Add back `winrtUninitialize` and deprecate it

halildurmus:
Added back the `winrtUninitialize()` helper function and deprecated it. Also updated some docs.

### DIFF
--- a/lib/src/winrt_helpers.dart
+++ b/lib/src/winrt_helpers.dart
@@ -12,7 +12,6 @@ import 'package:ffi/ffi.dart';
 
 import 'com/iinspectable.dart';
 import 'combase.dart';
-import 'constants.dart';
 import 'constants_nodoc.dart';
 import 'exceptions.dart';
 import 'guid.dart';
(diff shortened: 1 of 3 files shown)


lib/src/winrt_helpers.dart
```diff
@@ -27,9 +27,21 @@ import 'winrt/foundation/winrt_enum.dart';
...
+/// need your thread to be initialized with a single-threaded apartment (STA),
+/// you can call this function.
+///
 /// {@category winrt}
 void winrtInitialize() => RoInitialize(RO_INIT_TYPE.RO_INIT_SINGLETHREADED);
 
+/// Closes the Windows Runtime on the current thread.
+///
+/// {@category winrt}
+@Deprecated('Using this function may cause segfaults or runtime exceptions and '
```
timsneath: I think the text is a little too strong here (and quite scary!)

If I understood it correctly, the source of our intermittent flake is that in a tight loop of tests, `RoUninitialize` might still be executing when the next test is running from a different isolate (but shared thread). And as such the next call to initialize COM may hit a race condition. The new approach simply initializes COM on error, and relies on the process termination to tear it down. 

My proposal is that we deprecate both `winrtInitialize` and `winrtUninitialize`, with the following comments, respectively (needs better wrapping, probably!)

```dart
@Deprecated('winrtInitialize is no longer required. The Windows Runtime is '
  'automatically initialized by the Dart projection if it is not already initialized '
  'on a given thread when a WinRT class is activated. If you explicitly want to '
  'initialize the current thread with a specific threading model, use RoInitialize '
  'directly instead. This function will be removed in the next major release.')
void winrtInitialize() {}

@Deprecated('winrtUninitialize is not required in most scenarios, since '
  'Windows will clean up the process on exit. If you explicitly want to '
  'uninitialize the Windows Runtime, use RoUninitialize directly instead. '
  'This function will be removed in the next major release.')
void winrtUninitialize() {}
```

I'd advocate that `winrtUninitialize()` points to an empty function, since there's no harm in it being inactive.

lib/winrt.dart
```diff
@@ -19,11 +19,11 @@
...
 ///
-/// This package supports automatic initialization of the Windows Runtime on the
-/// current thread with a Multi-Threaded Apartment (MTA). However, if you wish
-/// to initialize the Windows Runtime manually or just call some APIs that only
-/// work in a Single-Threaded Apartment (STA), you can use the [winrtInitialize]
-/// helper function.
+/// This package ensures that threads are implicitly assigned to the
+/// multi-threaded apartment (MTA) so most of the time you don't need to do
+/// anything. However, if you need to use APIs that only work in a
+/// single-threaded apartment (STA), you need to call [winrtInitialize] helper
```
timsneath: Let's point users directly to `RoInitialize(RO_INIT_TYPE.RO_INIT_SINGLETHREADED)` rather than `winrtInitialize` so we can remove the latter in the future.

timsneath (changes requested):

lib/src/winrt_helpers.dart
halildurmus: > If I understood it correctly, the source of our intermittent flake is that in a tight loop of tests, `RoUninitialize` might still be executing when the next test is running from a different isolate (but shared thread). And as such the next call to initialize COM may hit a race condition. The new approach simply initializes COM on error, and relies on the process termination to tear it down.

Yes, I believe that's what happened.

Thanks, these are great!

lib/winrt.dart
halildurmus: Done

timsneath (approved):

PR #649, opened 2023-01-06
Convert acronyms in method/property names to lowercase

### LINKED ISSUE #648: Update WinRT method/property projection heuristics to match the Dart style guide
https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words

For example, this should be `ipInformation`:
https://github.com/timsneath/win32/blob/b720de0dcef968953816194783e0144f11386276/lib/src/winrt/networking/ihostname.dart#L39

halildurmus:
Fixes #648


timsneath (approved):
Awesome!

PR #650, opened 2023-01-06
Mark methods/properties that return WinRT objects as nullable

### LINKED ISSUE #647: Methods/properties that return WinRT objects should be nullable
For example, [NetworkInformation.GetInternetConnectionProfile Method](https://learn.microsoft.com/en-us/uwp/api/windows.networking.connectivity.networkinformation.getinternetconnectionprofile?view=winrt-22621#returns) states:
> The profile for the connection currently used to connect the machine to the Internet, or null if there is no connection profile with a suitable connection.

Here is an example that calls this method (make sure you are disconnected from the internet):
```dart
import 'package:win32/winrt.dart';

class ConnectionProfile extends IInspectable {
  ConnectionProfile.fromRawPointer(super.ptr);
}

ConnectionProfile getInternetConnectionProfile() {
  final activationFactoryPtr = CreateActivationFactory(
      'Windows.Networking.Connectivity.NetworkInformation',
      IID_INetworkInformationStatics);
  final object =
      INetworkInformationStatics.fromRawPointer(activationFactoryPtr);

  try {
    return ConnectionProfile.fromRawPointer(
        object.getInternetConnectionProfile());
  } finally {
    object.release();
  }
}

void main() {
  final profile = getInternetConnectionProfile();
  print(profile.runtimeClassName); // The program will exit with an Access Violation error code
}
```

There is no information on WinMD files on whether a method/property can return null or not. That's why we need to mark methods/properties that return WinRT objects (except async delegates e.g. `IAsyncAction`, collection interfaces e.g. `IVector`, `IMap` etc., and methods from `XXFactory` interfaces) as nullable (see https://github.com/microsoft/windows-rs/issues/331).

I actually did a similar thing in #465. FileOpenPicker's pickSingleFileAsync() returns `IAsyncOperation<StorageFile>`. Since the type argument is a WinRT object, we consider it nullable and expose it as `Future<StorageFile?>`. As pickMultipleFilesAsync() returns `IAsyncOperation<IVectorView<StorageFile>>` and the type argument is a collection interface we don't consider it nullable and expose it as `Future<List<StorageFile>>`.

halildurmus:
Fixes #647

Note that we must also mark WinRT object parameters as nullable. 

[IStorageItemExtraProperties.RetrievePropertiesAsync(IIterable<String>) Method](https://learn.microsoft.com/en-us/uwp/api/windows.storage.fileproperties.istorageitemextraproperties.retrievepropertiesasync?view=winrt-22621#parameters)'s parameter can take **null**:
> A collection that contains the names of the properties to retrieve. Pass null to retrieve all possible properties.

I'm thinking of doing that after we land this.

### DIFF
--- a/example/appcontainer.dart
+++ b/example/appcontainer.dart
@@ -43,6 +43,6 @@ void main() {
   print('${!isAppContainer() ? '!' : ''}isAppContainer');
 
   final userData = UserDataPaths.getDefault();
-  final roamingAppData = userData.roamingAppData;
+  final roamingAppData = userData?.roamingAppData;
   print('RoamingAppData: $roamingAppData');
 }
--- a/example/calendar.dart
+++ b/example/calendar.dart
@@ -19,7 +19,7 @@ void main() {
   final calendar = Calendar();
   print(calendarData(calendar));
 
-  final clonedCalendar = calendar.clone();
+  final clonedCalendar = calendar.clone()!;
   final comparisonResult = clonedCalendar.compare(calendar);
   print('Comparison result of calendar and its clone: $comparisonResult');
 
--- a/example/geolocation.dart
+++ b/example/geolocation.dart
@@ -21,11 +21,12 @@ void main() async {
     final location = await locator.getGeopositionAsync();
 
     if (location != null) {
-      print('Current location: (${location.coordinate.latitude}, '
-          '${location.coordinate.longitude})');
-      print('Location accuracy: ±${location.coordinate.accuracy} meters');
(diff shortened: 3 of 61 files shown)


timsneath (approved):
Approved with clenched teeth :) Here's hoping that in the future WinRT adds null safety.

halildurmus:
> Approved with clenched teeth :) 

Thanks! 

> Here's hoping that in the future WinRT adds null safety.

There was a proposal to introduce a `NotNull` attribute to the WinMD for this but apparently, they decided not to go with it (see https://github.com/microsoft/xlang/pull/708) :(.

PR #666, opened 2023-02-01
fix a dartdoc comment reference

devoncarew:
- fix a dartdoc comment reference

I know this isn't currently your biggest issue w/ dartdoc, but I saw this unresolved reference when playing with generation times and sizes locally.

```
Generating docs for library winrt from package:win32/winrt.dart...
  warning: unresolved doc reference [calloc.free]
    from winrt.Guid.toNativeGUID: (file:///Users/devoncarew/projects/timsneath/win32/lib/src/guid.dart:107:17)
```

### DIFF
--- a/lib/src/guid.dart
+++ b/lib/src/guid.dart
@@ -103,7 +103,7 @@ class Guid {
   /// location.
   ///
   /// It is the caller's responsibility to free the memory at the pointer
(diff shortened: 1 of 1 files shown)


timsneath (approved):
Thanks! I noticed that too, but you went beyond "noticing" to "fixing"!

Issue #669, opened 2023-03-01
Create a `LoggingAllocator`


timsneath:
A logging allocator class would behave as follows:

- For release builds, it would simply wrap the existing `calloc` / `free` model with no significant overhead
- For debug builds, it would track the number and size of allocations it had made, along with as much metadata as can be captured on the source of the allocation. On freeing memory, it would remove the equivalent entry from the tracking log. At any time, it should be possible to retrieve the log. 

The logging allocator would help detect memory leaks during test runs, by validating that all memory allocated had been freed by the end of the run. 

Idea inspired by @dcharkes.

halildurmus:
Yes please!

timsneath:
Potential solution: https://github.com/dart-lang/leak_tracker/issues/42

PR #677, opened 2023-04-06
Merge dev into main

timsneath:
@halildurmus Are we ready to review and potentially merge this? There is probably some back and forth between us, but I'd like to move forward if we can!


halildurmus:
As you mentioned [here](https://github.com/users/timsneath/projects/3/views/1?pane=issue&itemId=22289636), we need to decide whether to keep `IInspectable` in this package or not. I'm inclined towards keeping it here. 

If we're gonna keep it here, I'd like to propose some changes to both the `IInspectable` and `IUnknown` interfaces.

lib/src/com/iunknown.dart
```diff
@@ -46,7 +48,12 @@ class IUnknown implements Finalizable {
   ///
   /// If the COM object implements the interface, then it returns a pointer to
   /// that interface after calling `addRef` on it.
-  int queryInterface(Pointer<GUID> riid, Pointer<Pointer> ppvObject) => ptr
+  @Deprecated('Use toInterface instead. This function will be removed in the '
```
timsneath: I think I'd prefer not to deprecate this, but rather guide people towards `toInterface` as a helper function. The reason is consistency: it's an exported COM function (along with `addRef` and `release`), and I think it would be unique to actually remove it. In theory, someone else might build a COM helper library where they want to call it directly.

timsneath:
> As you mentioned [here](https://github.com/users/timsneath/projects/3/views/1?pane=issue&itemId=22289636), we need to decide whether to keep `IInspectable` in this package or not. I'm inclined towards keeping it here.
> 
> If we're gonna keep it here, I'd like to propose some changes to both the `IInspectable` and `IUnknown` interfaces.

My rationale for including it here is that it's a COM interface that is part of the Win32 metadata (namespace of `Windows.Win32.System.WinRT`). It's a peer to other Win32 APIs like `RoInitialize` and `WindowsCreateString` that sit at the boundary of WinRT. LMK if you have a different principle that you're applying.

lib/src/combase.dart
```diff
@@ -15,13 +15,19 @@ import 'macros.dart';
 import 'utils.dart';
 import 'win32/ole32.g.dart';
 
+/// Alias for a Pointer to COM vtable.
+typedef LPVTBL = Pointer<Pointer<IntPtr>>;
```
timsneath: I wonder if we should name this typedef `VTablePointer` (i.e. Dart-style casing)? Naming it in uppercase makes it look like this is a Win32 type like `LPWSTR` or `HANDLE`, when as far as I know it doesn't exist in windows.h.

lib/src/utils.dart
```diff
@@ -124,16 +124,15 @@ String convertFromHString(int hstring) =>
...
-    if (FAILED(hr)) {
-      throw WindowsException(hr);
-    } else {
-      return hString.value;
-    }
+    if (FAILED(hr)) throw WindowsException(hr);
+    return hString.value;
   } finally {
     free(stringPtr);
+    free(hString);
```
timsneath, line 135: Ohhh -- that's a brilliant catch. I missed that this is a Pointer<HSTRING> rather than the HSTRING itself.

halildurmus: Actually, my prototype for #669 caught this :). I've made some progress on #669 but let's finish restructuring Win32 / WinRT first.

lib/src/combase.dart
halildurmus: Makes sense, done.

lib/src/com/iunknown.dart
halildurmus: Done

halildurmus:
I removed IInspectable extension methods and added them as top-level functions instead (inspired by [How do I access and customize the IInspectable methods of a Windows Runtime class written in C++/WinRT?](https://devblogs.microsoft.com/oldnewthing/20220421-00/?p=106511)).

halildurmus (approved):
LGTM

PR #691, opened 2023-05-06
Switch to `Finalizer`

halildurmus:
- Switches to [Finalizer](https://api.dart.dev/stable/3.0.0/dart-core/Finalizer-class.html) in `IUnknown` to call `.release()` automatically when objects go out of scope
- Removes `Unloading COM support` section in `doc/com.md`
I'm worried that in some cases, if `.release()` is called (by Finalizer) after COM is unloaded due to call to `CoUninitialize`, it'll lead to use after free that causes the process to crash, so I think it's better to let COM torn down when the process is terminated (we did something similar for WinRT and removed the mention of `winrtUninitialize` which was calling `RoUninitialize` #642)
- Updates examples and tests accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## 5.0.0
+
+- Added `detach` method to `IUnknown`, which detaches the object from
+  `Finalizer` if you want to manually manage the lifetime of the object
+  yourself (#677)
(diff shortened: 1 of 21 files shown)


timsneath (approved):
Awesome! Thanks!

PR #692, opened 2023-05-06
Upgrade to support Dart 3

timsneath:


timsneath:
We can merge this with your review, @halildurmus. I'm not worried about the `stable` channel right now; it'll be running Dart 3 soon enough :)

halildurmus (approved):
LGTM 🎉

PR #712, opened 2023-05-27
Update docs.microsoft.com URLs

loic-sharma:
Updates doc.microsoft.com URLs:

1. **Removes language codes** - Removing the `en-us` language code from the URL's path lets the docs website redirect to whichever locale is most appropriate for the user.
2. **Update to the 'learn' subdomain** - The `docs.microsoft.com` subdomain has been replaced with `learn.microsoft.com`.


halildurmus (approved):
Thanks!

PR #715, opened 2023-06-01
Create iwinhttprequest

lublak:
https://github.com/dart-windows/win32/issues/714

Currently i have no idea how to find the right index for elementAt (i dont think they are correct in my code)
and i have no idea how to use Long and VARIANT_Bool.


halildurmus:
Superseded by #718

PR #752, opened 2023-09-19
Set `guidItem` in `NOTIFYICONDATA` struct

### LINKED ISSUE #674: shell_notify_icon example does not work
image

It seems that shell_notify_icon example does not longer work on latest windows versions, i've tried to launch with both dart run and flutter run, there is no icon in tray and app stucks

halildurmus:
Fixes #674


domesticmouse (approved):

domesticmouse:
Can you configure things so CI is required to pass before automatically merging?

halildurmus:
Done. Thanks for catching this.

PR #800, opened 2023-12-12
Add more WASAPI functions, interfaces and constants

### LINKED ISSUE #513: Missing property : PROPVARIANT.pwszVal
I am trying to get audio devices names but apparently pwszVal is missing from PROPVARIANT or it is set somewhere else. 
Here is [c++ function to get audio devices names](https://github.com/pauldotknopf/WindowsSDK7-Samples/blob/3f2438b15c59fdc104c13e2cf6cf46c1b16cf281/multimedia/audio/EndpointVolume/endpointvolumechanger.cpp#L67)

What should I do now?

My code:
```dart


import 'dart:ffi';
import 'package:win32/win32.dart';
import 'package:ffi/ffi.dart';
import './com.immdevicecollection.dart';
import './com.ipropertystore.dart';

void check(int hr) {
}

void main() {
  check(CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED));

  final pDeviceEnumerator = MMDeviceEnumerator.createInstance();

  //Enum Devices
  final deviceEnumPoints = calloc<Pointer<COMObject>>();
  pDeviceEnumerator.EnumAudioEndpoints(0, 1, deviceEnumPoints);
  IMMDeviceCollection collection = IMMDeviceCollection(deviceEnumPoints.cast());
  free(deviceEnumPoints);

  //Count Devices
  final cCountItems = calloc<Uint32>();
  collection.GetCount(cCountItems);
  final items = cCountItems.value;
  free(cCountItems);

  //Loop through each
  for (int i = 0; i < items; i++) {
    //Get Device
    final pDevice = calloc<Pointer<COMObject>>();
    collection.Item(i, pDevice);
    final pDevice2 = IMMDevice(pDevice.cast());
    free(pDevice);

    //get Properties
    final pProperties = calloc<Pointer<COMObject>>();
    pDevice2.OpenPropertyStore(STANDARD_RIGHTS_READ, pProperties);
    IPropertyStore props = IPropertyStore(deviceEnumPoints.cast());
    free(pProperties);
    //Count them
    final cProps = calloc<UINT32>();
    props.GetCount(cProps);

    //PROPVARIANT
    final guid = calloc<GUID>();
    var pKey = calloc<PROPERTYKEY>();
    var pValue = calloc<PROPVARIANT>();
    CLSIDFromString("{a45c254e-df1c-4efd-8020-67d146a850e0}".toNativeUtf16(), guid);
    pKey.ref.fmtid = guid.ref;
    pKey.ref.pid = 14;
    props.GetValue(pKey, pValue);
    final out = "${pValue.ref.val1} ${pValue.ref.val2} ${pValue.ref.vt} ${pValue.ref.wReserved1} ${pValue.ref.wReserved2} ${pValue.ref.wReserved3}";
    print(out);

    free(guid);
    free(pKey);
    free(pValue);
    free(cProps);

  }
  print("Total Items: $items");

  pDeviceEnumerator.Release();
  free(pDeviceEnumerator.ptr);
}
```

New interfaces: 
```json

  'Windows.Win32.Media.Audio.IMMDeviceCollection',
  'Windows.Win32.UI.Shell.PropertiesSystem.IPropertyStore'
```

Generated classes here:

```dart
import 'dart:ffi';

import 'package:win32/win32.dart';
import 'package:ffi/ffi.dart';

/// @nodoc
const IID_IMMDeviceCollection = '{0BD7A1BE-7A1A-44DB-8397-CC5392387B5E}';

/// {@category Interface}
/// {@category com}
class IMMDeviceCollection extends IUnknown {
  // vtable begins at 3, is 2 entries long.
  IMMDeviceCollection(super.ptr);

  int GetCount(Pointer<Uint32> pcDevices) => ptr.ref.vtable
      .elementAt(3)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Pointer<Uint32> pcDevices)>>>()
      .value
      .asFunction<int Function(Pointer, Pointer<Uint32> pcDevices)>()(ptr.ref.lpVtbl, pcDevices);

  int Item(int nDevice, Pointer<Pointer<COMObject>> ppDevice) => ptr.ref.vtable
      .elementAt(4)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Uint32 nDevice, Pointer<Pointer<COMObject>> ppDevice)>>>()
      .value
      .asFunction<int Function(Pointer, int nDevice, Pointer<Pointer<COMObject>> ppDevice)>()(ptr.ref.lpVtbl, nDevice, ppDevice);
}

const IID_IPropertyStore = '{886D8EEB-8CF2-4446-8D02-CDBA1DBDCF99}';

/// {@category Interface}
/// {@category com}
class IPropertyStore extends IUnknown {
  // vtable begins at 3, is 5 entries long.
  IPropertyStore(super.ptr);

  int GetCount(Pointer<Uint32> cProps) => ptr.ref.vtable
      .elementAt(3)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Pointer<Uint32> cProps)>>>()
      .value
      .asFunction<int Function(Pointer, Pointer<Uint32> cProps)>()(ptr.ref.lpVtbl, cProps);

  int GetAt(int iProp, Pointer<PROPERTYKEY> pkey) => ptr.ref.vtable
      .elementAt(4)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Uint32 iProp, Pointer<PROPERTYKEY> pkey)>>>()
      .value
      .asFunction<int Function(Pointer, int iProp, Pointer<PROPERTYKEY> pkey)>()(ptr.ref.lpVtbl, iProp, pkey);

  int GetValue(Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> pv) => ptr.ref.vtable
      .elementAt(5)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> pv)>>>()
      .value
      .asFunction<int Function(Pointer, Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> pv)>()(ptr.ref.lpVtbl, key, pv);

  int SetValue(Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> propvar) => ptr.ref.vtable
      .elementAt(6)
      .cast<Pointer<NativeFunction<Int32 Function(Pointer, Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> propvar)>>>()
      .value
      .asFunction<int Function(Pointer, Pointer<PROPERTYKEY> key, Pointer<PROPVARIANT> propvar)>()(ptr.ref.lpVtbl, key, propvar);

  int Commit() => ptr.ref.vtable.elementAt(7).cast<Pointer<NativeFunction<Int32 Function(Pointer)>>>().value.asFunction<int Function(Pointer)>()(ptr.ref.lpVtbl);
}
```

SirusDoma:
Add more WASAPI functions, interfaces, and constants.
Fix: https://github.com/dart-windows/win32/issues/513


example/wasapi.dart
```diff
@@ -83,8 +83,68 @@ void main() {
...
+        STGM.STGM_READ, // Storage-access mode: read
+        ppProps.cast()));
+
+    // Build property key to get device friendly name.
+    final pProps = IPropertyStore(ppProps.cast());
+    final guid = calloc<GUID>();
+    CLSIDFromString(PKEY_DeviceInterface_FriendlyName.toNativeUtf16(), guid);
+    final pKey = calloc<PROPERTYKEY>()
+      ..ref.fmtid = guid.ref
+      ..ref.pid = 2;
```
SirusDoma: The annoying thing about getting the property value from property store is that we need to use the correct `GUID` **AND** Property ID (`pid`). Specifying both every `pProps.getValue` calls looks very error-prone to me.  

Is there a way to couple these `GUID` and `pid` together? If so, how should I implement it? An enum in some place perhaps?

SirusDoma: Just for reference, the C# binding for `win32` seems to store these guids inside a static class and couple the guid and pid together via a C# attribute called `Constant`.

I do think we should do something like this as well.
image

SirusDoma:
A bit side-tracked but after testing these API in some of my small projects, I ran into an issue when using `IAudioCaptureClient`.
Quoting from [MSDN docs](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nn-audioclient-iaudiocaptureclient):

> When releasing an IAudioCaptureClient interface instance, the client must call the Release method of the instance from the same thread as the call to IAudioClient::GetService that created the object.

I couldn't find a way to do this in Flutter. Every time the code executes from a button click, the 2 different methods of the same object seem to run under a different native thread leading to segfault / silent crash. My temporary solution is to let this capture client pointer leaking to avoid the crash.. Which is far more than ideal.

Any ideas on how to do this? I reckon I could use a native thread via `CreateThread` and call the `GetService` and  `Release` under the same thread, but that would be complicated to do.

example/wasapi.dart
halildurmus: Is this screenshot from ILSpy? Just to clarify, it doesn't represent the C# bindings for the `PROPERTYKEY` properties. ILSpy provides a glimpse into the API metadata being introspected. It's up to language projection authors to decide how to utilize this information.

[Constant](https://github.com/microsoft/win32metadata/blob/bcce701782815f62c550465072c66592e330ff35/generation/WinSDK/manual/Metadata.cs#L66-L71) attribute is part of the [win32metadata](https://github.com/microsoft/win32metadata) and it includes values of the struct initializers as a comma-separated string. In this case, `PROPERTYKEY`'s `fmtid` and `pid` values.

For example, you can check out the Rust projection of these properties [here](https://github.com/microsoft/windows-rs/blob/02db74cf5c4796d970e6d972cdc7bc3967380079/crates/libs/windows/src/Windows/Win32/Devices/FunctionDiscovery/mod.rs#L1187-L1189), which is auto-generated from the win32metadata by utilizing the values inside the `Constant` attribute.

In Rust, destructors can be implemented for any struct through the Drop trait. This is typically done to ensure the automatic cleanup of resources when an instance of the struct goes out of scope. Such a mechanism enables the creation of constants, like the ones illustrated. However, Dart does not support this. Consequently, manual resource deallocation is required, rendering it impractical to create constants or enums for properties of this nature in Dart.

BTW, there is a better way to create these properties in your code.

Instead of doing this:
```dart
  final guid = calloc<GUID>();
  CLSIDFromString(PKEY_DeviceInterface_FriendlyName.toNativeUtf16(), guid);
  final pKey = calloc<PROPERTYKEY>()
    ..ref.fmtid = guid.ref
    ..ref.pid = 2;

  ...

  free(guid);
  free(pKey);
```

You can do this:
```dart
  final PKEY_DeviceInterface_FriendlyName = calloc<PROPERTYKEY>()
    ..ref.fmtid.setGUID('{026e516e-b814-414b-83cd-856d6fef4822}')
    ..ref.pid = 2;

  ...

  // Free the memory allocated for the struct when it is no longer needed
  free(PKEY_DeviceInterface_FriendlyName);
```

halildurmus:
> A bit side-tracked but after testing these API in some of my small projects, I ran into an issue when using `IAudioCaptureClient`. Quoting from [MSDN docs](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nn-audioclient-iaudiocaptureclient):
> 
> > When releasing an IAudioCaptureClient interface instance, the client must call the Release method of the instance from the same thread as the call to IAudioClient::GetService that created the object.
> 
> I couldn't find a way to do this in Flutter. Every time the code executes from a button click, the 2 different methods of the same object seem to run under a different native thread leading to segfault / silent crash. My temporary solution is to let this capture client pointer leaking to avoid the crash.. Which is far more than ideal.
> 
> Any ideas on how to do this? I reckon I could use a native thread via `CreateThread` and call the `GetService` and `Release` under the same thread, but that would be complicated to do.

[Finalizer](https://api.dart.dev/stable/dart-core/Finalizer-class.html)s are attached to the COM objects in this package, which decrement the reference count of the object and then free its pointer (see [here](https://github.com/dart-windows/win32/blob/b27171af8179465b41d33a440e7c0c70310a515c/lib/src/com/iunknown.dart#L30-L39)) when the objects go out of scope (as mentioned in the [docs](https://win32.pub/docs/com-programming/basics#releasing-com-objects)). Therefore, you should not free the pointers of these objects unless you're opting out of this behavior by calling the [.detach()](https://pub.dev/documentation/win32/latest/win32/IUnknown/detach.html) method of the object.

Edit: I'll review this PR in a few days, I'm a little busy at the moment.

SirusDoma:
> [Finalizer](https://api.dart.dev/stable/dart-core/Finalizer-class.html)s are attached to the COM objects in this package, which decrement the reference count of the object and then free its pointer (see [here](https://github.com/dart-windows/win32/blob/b27171af8179465b41d33a440e7c0c70310a515c/lib/src/com/iunknown.dart#L30-L39)) when the objects go out of scope (as mentioned in the [docs](https://win32.pub/docs/com-programming/basics#releasing-com-objects)). Therefore, you should not free the pointers of these objects unless you're opting out of this behavior by calling the [.detach()](https://pub.dev/documentation/win32/latest/win32/IUnknown/detach.html) method of the object.

Yes, I'm completely aware of what you wrote here. However, it doesn't matter if I let the finalizer run (e.g class that contains the `IAudioCaptureClient` / COM object went out of scope) or detach the pointer from the flutter object and freeing the object by myself: It always leads to segfault/crash because I think internally runs in a different **native** thread than the one it used to instantiate the object.

The problem isn't likely to be reproducible by example code where everything happens in a single method (eg. init -> render/capture -> object being freed due to out-of-scope at the same method), but it happens to me in a flutter project where init, capture and free/dispose happen within different methods (e.g `initState()` for `init`, `capture` after user-click and/or dispose when widget that contains the COM object no longer active / being rendered / out-of-scope).

> Edit: I'll review this PR in a few days, I'm a little busy at the moment.

Please take your time, I'm using my local build for my project for now :)

halildurmus:
> The problem isn't likely to be reproducible by example code where everything happens in a single method (eg. init -> render/capture -> object being freed due to out-of-scope at the same method), but it happens to me in a flutter project where init, capture and free/dispose happen within different methods (e.g `initState()` for `init`, `capture` after user-click and/or dispose when widget that contains the COM object no longer active / being rendered / out-of-scope).

If you can share an example of this, I'd be happy to look into it.

example/wasapi.dart
SirusDoma: Hi, I've updated the code as your suggestion. And yes, it is a similar program to ILSpy called dnSpy. I see, it would make sense for Rust, but I believe my point still stand. Manually creating `PROPERTYKEY` by setting GUID and pid is sounds very error-prone to me, especially when the program deals with many properties.  

Not to mention, looking for the GUID and PID of these property keys is somewhat not trivial. The MSDN docs doesn't mention these values. I had to head to https://www.magnumdb.com/ to see the values.

I would love to see a way to at least couple these GUID and its pid in one place. I was thinking about creating an enum that consists only of the GUID string and PID and then let the user construct the `PROPERTYKEY` by themselves with these values enum values.

SirusDoma:
> If you can share an example of this, I'd be happy to look into it.

Noted! I'll post a small reproducible flutter code/project later.

example/wasapi.dart
halildurmus: We can consider adding static methods into the `PROPERTYKEY` struct:

```dart
base class PROPERTYKEY extends Struct {
  external GUID fmtid;

  @Uint32()
  external int pid;

  static Pointer<PROPERTYKEY> DeviceInterface_FriendlyName({
    Allocator allocator = calloc,
  }) =>
      allocator<PROPERTYKEY>()
        ..ref.fmtid.setGUID('{026e516e-b814-414b-83cd-856d6fef4822}')
        ..ref.pid = 2;

  ...
}
```

SirusDoma:
Hi, I'm so sorry that I haven't updated much yet. I've been busy lately.  
Anyway, about the issue about releasing the COM Object, I think the issue lies in my capture codes and not in the flutter or thread stuff. I'm not absolutely certain and have yet to confirm this. I'll post the update once I find something.

P.S I liked your commit [Add static methods to PROPERTYKEY struct](https://github.com/dart-windows/win32/pull/800/commits/6874df288cd06a99bc7a6fdda3ac1ef30f258d38), not sure why you revert it.

halildurmus:
> P.S I liked your commit [Add static methods to PROPERTYKEY struct](https://github.com/dart-windows/win32/pull/800/commits/6874df288cd06a99bc7a6fdda3ac1ef30f258d38), not sure why you revert it.

The commit caused some [failures](https://github.com/dart-windows/win32/actions/runs/7262579394) that I haven't been able to diagnose. I'll make another attempt tomorrow to identify and address the issues.

halildurmus (approved):
LGTM, thank you for your contribution!

I removed the `ActivateAudioInterfaceAsync` function and related COM interfaces as supporting asynchronous Win32 APIs requires some extra work and I don't have any plans to tackle this at the moment.

SirusDoma:
Sorry to bring this up again. There's no problem with releasing the COM object. It appears I'm using `Isolate` incorrectly.  
Anyway, when is this going to be released?

